PR #2, opened 2023-07-27
resources: Add a tutorial on adding workloads

kunpai:
This document details how to add a Resource, wrap it in a Workload, and test the Workload with the new gem5 Resources infrastructure introduced in v23.0.


ADDING-WORKLOADS.md
```diff
@@ -0,0 +1,70 @@
...
+    ],
+    "resources": {
+        "binary": "binary-resource"
+    },
+}
+```
+
+## Testing the Workload
+
+To test the Workload, we first have to add the local JSON file as a data source for gem5. This can be done by navigating to `src/python/gem5_default_config.py` and adding a JSON object to `"sources"` in the following format:
```
powerjg: I really don't like this suggestion. We should not require people to modify something in source and recompile gem5 just to test a local workload.

Do we not have a way to use an environment variable to point to a json file? If not, we should add that!

kunpai: According to this function we wrote: [https://github.com/gem5/gem5/blob/af72b9ba580546ac12ce05bfaac3fd53fa8699f4/src/python/gem5/resources/client.py#L53](https://github.com/gem5/gem5/blob/af72b9ba580546ac12ce05bfaac3fd53fa8699f4/src/python/gem5/resources/client.py#L53), we can either place a gem5 resources config file in the cwd or pass its path in through GEM5_CONFIG while building. So, I should indicate those two options?

powerjg: Yes! Thanks :)

powerjg (dismissed):

ADDING-WORKLOADS.md
```diff
@@ -0,0 +1,78 @@
...
+    "id": "binary-resource",
+    "description": "A RISCV binary used to test a specific RISCV instruction.",
+    "architecture": "RISCV",
+    "is_zipped": false,
+    "resource_version": "1.0.0",
+    "gem5_versions": [
+        "23.0"
+    ],
+}
+```
```
BobbyRBruce, line 42: How is the user supposed to set the path to the the binary resource in this case?

ADDING-WORKLOADS.md
```diff
@@ -0,0 +1,78 @@
+# Adding a Workload to gem5
+
+This tutorial will walk you through the process of creating a Workload in gem5 and testing it, through the new gem5 Resources infrastructure introduced in gem5 v23.0.
+
+## Introduction
```
BobbyRBruce, line 14: I would explain what a workload is. Particularly that it declaring a function call to a board.

`board.set_workload('x')` -> this. The Workload with ID 'x' will be parsed and it will be used to construction the function call it defines.

ADDING-WORKLOADS.md
```diff
@@ -0,0 +1,78 @@
...
+
+Assuming you have the Resource JSON and the Resource is part of gem5 Resources, you can now define the Workload JSON. Let's assume that the Workload we are building wraps `binary-resource`, and is called `binary-workload`. We can define this Workload in a local JSON file as follows:
+
+``` json
+{
+    "id": "binary-workload",
+    "category": "workload",
+    "description": "A RISCV binary used to test a specific RISCV instruction.",
+    "architecture": "RISCV",
+    "function": "set_se_binary_workload",
```
BobbyRBruce, line 56: I'd link to the `set_se_binary_workload` function here and explain the `resources` field here specifies the setting of the binary. You should also mention that the `additional_parameters` field. Why not have this binary accept some arguments?

BobbyRBruce (changes requested):

ADDING-WORKLOADS.md
kunpai: I can look into this, but my purpose of not mentioning it was also on the base assumption that the Resource should exist on gem5 Resources before one makes a Workload of it. Should we allow both to be done simultaneously?

With the REU cohort, they are making Workloads out of Resources that already exist or are making the Resource first before they make a Workload, which is why the tutorial follows its current flow.

ADDING-WORKLOADS.md
kunpai: I like this suggestion!

ADDING-WORKLOADS.md
kunpai: Sounds good! I'll work on it.

ADDING-WORKLOADS.md
BobbyRBruce: I don't think this makes much sense. Perhaps I'm missing something. If you define the resource inside the local JSON file then it'll use that resource. In this case `binary-resource`. That resource is not in the database. How is gem5 supposed to know where this binary is?

powerjg (changes requested):
Before merging, can you clean up the git history (i.e., rebase and/or squash) into a human-understandable history?

BobbyRBruce:
Did you mean to close this?

kunpai:
Yes, I need to edit the tutorial. I will reopen it once Workloads are deprecated and my change on supporting paths is merged

PR #6, opened 2023-08-16
resources: Add x86-gapbs on Ubuntu 22.04 disk image resource

erin-le:
This commit adds the files necessary to create an x86 Ubuntu 22.04 disk image containing the GAP Benchmark Suite. It only includes the files necessary to create the disk image, and not the disk image itself. The files in the directory "qemu_files" are used to create the disk image and launch a QEMU instance. The directory "packer_files" contains the files needed by Packer to install GAPBS onto the disk image. The README contains detailed instructions on how to build the disk image, and x86-gapbs-benchmarks-2204.py is a sample script.

After the disk image is built, it can be used as a DiskImageResource in a gem5 simulation.


powerjg:
@hnpl your review would be helpful!

@erin-le what testing have you done?

Finally, @KyleRoarty you mentioned that you figured out a way to build this without  separately launching qemu (I think). Could you let us know how you did that (and if you could provide a review that could be helpful as well).

erin-le:
I tested different combinations of sizes and benchmarks for the original disk image I made. After testing the original disk image, I wrote instructions to make the disk image, remade the disk image based on those, and ran a few simulations to check that the benchmarks would run and the simulation would finish properly. I did this again after editing the instructions and adding a few files.
I used synthetic graphs with sizes between 1 and 16, as well as the USA-road-d.NY.gr real graph. The benchmarks I tested were cc,bc,tc,pr,and bfs. For each benchmark, I tried the following sizes on the original disk:
bc: USA-road, 2, 4, 8, 10, 14
bfs: USA-road, 4, 8, 16
cc: USA-road, 2, 3, 10
pr: 2, 5, 6, 7, 8, 9, 12, 14, 16
tc: 2, 7, 10, 12, 16

src/x86-gapbs-ubuntu-22.04/packer_files/1.packages-install.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Removing mounting /boot/efi
+sudo sed -i '/\/boot\/efi/d' /etc/fstab
+sudo systemctl stop boot-efi.mount
+sudo systemctl disable boot-efi.mount
+
+# Removing cloud-init
+sudo touch /etc/cloud/cloud-init.disabled
+
+# Use `systemctl mask <service/target/socket/timer>` to stop a systemd object from starting during the booting time
```
hnpl: The line 28-30 are comments for trimming systemd services. Since we are not removing systemd services, can you remove the comments?

src/x86-gapbs-ubuntu-22.04/qemu_files/cloud.txt
```diff
@@ -0,0 +1,9 @@
+#cloud-config
+users:
+  - name: ubuntu
+    lock_passwd: false
+    groups: sudo
+    sudo: ['ALL=(ALL) NOPASSWD:ALL']
+    shell: /bin/bash
+    ssh-authorized-keys:
+      - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDg18gIYyZcGaAbhgZ1d20b0Id2AKIHxesNRFCoaJmrkAZfxbXJ6wpybtyLJizISQUoSoWugGThg25/Hkb8WP9CNvn4R/CJWnzB1TKblY5WmeTDDr0DDtl+2dtaZOTBccCXAWSyPVT3fmMmkxmTY/Ttz2DwDsZg5c+L1hLgRcL6nTLBcpp1qsj/O8czcumIptz/RY98Y9bVyM8RMA/b3KuUB94kXHocZQ8gFJFPTete5CJ15tEliBHRc7ugJ+FXVWLBWEKUpTdb/Z5kkfXeKwPWeihG//b7ejABuODsv9Jg9OwsBqqKH8TibRFiNJfDYORC52d9O983Hm50jPMop0GFsb8lQYiiS3b3+4rllDJwJKR8wAENCBz3Gj66EIj2HO79EhaDpWgtcLHLknqZgnQ6RpIwta0YcISH7hq5laJEq/jC+jGnKf9A9NMf5dEIHN16zHqsXHQApiAcBocIgoyx3HuoLUKl/6pWe504PjreTcVTbn/77zrCK9EsNJPRgT0= ubuntu@localhost
```
hnpl: This key should be generated by user. Reusing public keys could be a security issue.

src/x86-gapbs-ubuntu-22.04/qemu_files/get-disk-img.sh
```diff
@@ -0,0 +1,3 @@
+wget https://cloud-images.ubuntu.com/releases/22.04/release-20230616/ubuntu-22.04-server-cloudimg-amd64.img
+qemu-img convert ubuntu-22.04-server-cloudimg-amd64.img -O raw ./x86_64-hpc-2204.img
+qemu-img resize -f raw ./x86_64-hpc-2204.img +20G
```
hnpl: This line adds about 20GiB of disk space to the disk, which is rather large if someone wants to download the disk image. User can always increase the disk size, so I think it's safe to minimize the disk size. I'm not sure what is the optimal size, but can you test with +3GiB or +5GiB?

src/x86-gapbs-ubuntu-22.04/qemu_files/flash0.img
hnpl, line 1: I don't think we need to upload this file as the file can be generated by a script or by user.

src/x86-gapbs-ubuntu-22.04/qemu_files/flash1.img
hnpl, line 1: I don't think we need to upload this file as the file can be generated by a script or by user.

src/x86-gapbs-ubuntu-22.04/packer_files/post-installation.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Copyright (c) 2020 The Regents of the University of California.
+# SPDX-License-Identifier: BSD 3-Clause
+
+
+echo 'Post Installation Started'
+
+mv /home/ubuntu/serial-getty@.service /lib/systemd/system/
+
+# mv /home/ubuntu/gem5/m5 /sbin
```
hnpl: Can you remove the line 11-12 since they are commented out?

src/x86-gapbs-ubuntu-22.04/packer_files/2.m5-install.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Copyright (c) 2023 The Regents of the University of California.
+# SPDX-License-Identifier: BSD 3-Clause
+
+cd $HOME
+git clone https://gem5.googlesource.com/public/gem5/
```
hnpl: Can you change this link to https://github.com/gem5/gem5

src/x86-gapbs-ubuntu-22.04/x86-gapbs-benchmarks-2204.py
```diff
@@ -0,0 +1,290 @@
...
+    switch_core_type=CPUTypes.TIMING,
+    #switch_core_type=CPUTypes.O3,
+    isa=ISA.X86,
+    num_cores=2,
+)
+
+# Here we setup the board. The X86Board allows for Full-System X86 simulations
+
+
+# class X86Board_sda(X86Board):
```
hnpl, line 182: Can you add comments to the script explaining when and how to use line 179-189?

src/x86-gapbs-ubuntu-22.04/qemu_files/cloud.img
hnpl, line 1: I don't think we need to upload this file as the file can be generated by a script or by user.

src/x86-gapbs-ubuntu-22.04/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+(3) You will also need to either (a)generate a new pair of ssh keys, replace the public key in `cloud.txt`, and change the filepath of the private key in `x86_64-hpc.json`, or (b)generate a new private key from the provided public key and update the filepath in `x86-64-hpc.json`. 
+    - The private key filepath in `x86-64-hpc.json` is under "ssh_certificate_file", in "builders"
+    - The public key in `cloud.txt` is under ssh-authorized-keys, at the bottom of the file.
+
+(4, 5) After this, make cloud.img by running `make-cloud-img.sh`and launch the disk image with `qemu-launch.sh`.
+
+(6) After launching the qemu instance, run Packer (in a different terminal) using `./packer build x86_64-hpc.json` in the packer_files directory. 
+
+Please note that the `x86_64-hpc.json` included in packer_files/x86_64-hpc.json is a modified version of the one in the linked tutorial. All scripts except for the first two were removed, as they weren't necessary for the GAPBS benchmarks, and `gapbs-install.sh` and `post-installation.sh` were added.
```
hnpl:
```suggestion
Please note that the `x86_64-hpc.json` included in `packer_files/x86_64-hpc.json` is a modified version of the one in the linked tutorial. All scripts except for the first two were removed, as they weren't necessary for the GAPBS benchmarks, and `gapbs-install.sh` and `post-installation.sh` were added.
```

hnpl (commented):
There are a few comments to be addressed but this looks good to me.

src/x86-gapbs-ubuntu-22.04/x86-gapbs-benchmarks-2204.py
```diff
@@ -0,0 +1,290 @@
+# Copyright (c) 2021 The Regents of the University of California.
```
BobbyRBruce: The copyright here is out-of-date.

BobbyRBruce: Why are you adding an example script here. Shouldn't this be in "configs/examples/gem5_library/" in the main gem5 repo?

BobbyRBruce (changes requested):
A few high-level comments here:

1. Perhaps I've missed it somewhere, but where do you add the ExitEvents? What have you addeded and where?
2. Should this be an update of the current x86-gapbs disk image (e.g., a Version 2 of it) or should this be a whole different resources? It seems to me like the former is correct.
3. Is the true intention here for this to be on the develop branch? If these changes are compatible with v23.0 then they should be on stable.

src/x86-gapbs-ubuntu-22.04/qemu_files/flash0.img
kroarty-lanl: This isn't used in the qemu script, so it can be removed

src/x86-gapbs-ubuntu-22.04/qemu_files/flash1.img
kroarty-lanl: ditto to flash0.img

src/x86-gapbs-ubuntu-22.04/qemu_files/qemu-launch.sh
```diff
@@ -0,0 +1,5 @@
+qemu-system-x86_64 \
+     -nographic -m 16384 -smp 8 \
+     -device virtio-net-pci,netdev=eth0 -netdev user,id=eth0,hostfwd=tcp::5558-:22 \
+     -drive file=x86_64-hpc-2204.img,format=raw \
+     -drive if=none,id=cloud,file=cloud.img -device virtio-blk-pci,drive=cloud
```
kroarty-lanl, line 5: It's possible to avoid compiling the cloud.txt into an img and passing it into qemu. [This](https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html) is the doc page explaining it.

I've only used the http method, so I can't verify exactly how to make the file method work, but cloud.txt would need to be renamed to user-data, a meta-data file would need to be created (ie `touch meta-data`) and this line would be replaced with something like `-smbios type=1,serial=ds=nocloud;s=file://path/to/gem5-resources/src/x86-gapbs-ubuntu-22.04/qemu-files/`

Given that I'm not sure how the file path works (does it have to be absolute) there might not be value in this change as the users may still have to interact (execute shell script vs put their path in).

src/x86-gapbs-ubuntu-22.04/qemu_files/cloud.txt
kroarty-lanl: You can use a password instead of a ssh key by setting the top-level key `ssh_pwauth: true`

If you decide to do this, you can then replace `ssh-authorized-keys` with `plain_text_passwd: <password>`

Also if this change is made, make sure to verify that the packer install files still work. I know the `sudo: ['ALL=(ALL) NOPASSWD:ALL']` line is supposed to allow sudo without using a password, but I ran into some issues (which may just be me messing up) and ended up piping the password into sudo.

src/x86-gapbs-ubuntu-22.04/packer_files/x86_64-hpc.json
```diff
@@ -0,0 +1,51 @@
...
+    "_author": "Hoa Nguyen <hoanguyen@ucdavis.edu>",
+    "_license": "Copyright (c) 2023 The Regents of the University of California. SPDX-License-Identifier: BSD 3-Clause",
+    "builders":
+    [
+        {
+            "type": "null",
+            "ssh_host": "localhost",
+            "ssh_port": "5558",
+            "ssh_username": "{{ user `ssh_username` }}",
+            "ssh_agent_auth": true,
```
kroarty-lanl, line 11: If using ssh_pwauth, remove lines 11-14, add an ssh_password variable to variables, and add in a line here 
```
"ssh_password": "{{ user `ssh_password` }}"
```

src/x86-gapbs-ubuntu-22.04/qemu_files/qemu-launch.sh
kroarty-lanl: Additional comment: Is this going to use kvm by default? If not, would it be worth adding to the qemu command?

kroarty-lanl (commented):
This is Kyle under my work account, Jason asked me to review this. 

As Jason mentioned, I've got a way to encapsulate this all into a packer script, but we determined it'd be better to get this checked in and I'll provide a pull request to update this to the format I use.

I have some feedback that could either be implemented here, or could wait until after when I submit my PR.

erin-le:
> 1. Perhaps I've missed it somewhere, but where do you add the ExitEvents? What have you addeded and where?
>2. Should this be an update of the current x86-gapbs disk image (e.g., a Version 2 of it) or should this be a whole different resources? It seems to me like the former is correct.
>3. Is the true intention here for this to be on the develop branch? If these changes are compatible with v23.0 then they should be on stable.

Responses to Bobby's review:
1. I'm not entirely sure. I think there's an ExitEvent added in `packer_files/gem5-init.sh` (a script that is run on the disk image), where the simulation exits after the benchmark finishes on the disk image, but I could also be misunderstanding what an ExitEvent is.
2. It looks like the intention was for this to be an update of the current x86-gapbs disk image. In this case, should I put the added `x86-gapbs-ubuntu-22.04` directory under the existing `x86-ubuntu` directory and group the files currently in the `x86-ubuntu` directory into a sub-directory?
3. I mainly made the pull request on the develop branch because of the instructions in the README. Would you want me to switch it to the stable branch?

src/x86-gapbs-ubuntu-22.04/x86-gapbs-benchmarks-2204.py
erin-le: I wasn't really sure where to put it. In that case, should I remove the file and make a new PR for it in the gem5 repo?

abmerop:
Is this something that could be / is intended to be a generic X86 disk image?  I would like to update the GPU disk image to 22.04, but what I want to do is start with a generic X86 disk and install the additional packages to a secondary disk rather than having multiple GBs of redundant files.

PR #8, opened 2023-09-05
Resources: ARM64 Ubuntu 22.04 Disk Image

erin-le:
This adds the files needed to create an ARM64, Ubuntu 22.04 disk image for use as a gem5 resource. The directory qemu_files contains the scripts for obtaining the disk image and launching it as a QEMU instance. The directory packer_files contains the files for obtaining Packer and setting up the disk image. README.md contains detailed instructions on how to build the disk image using these files.

Additionally, I looked at the comments that Hoa made on the X86 GAPBS pull request and checked them for this disk image as well. The disk image works when +3G is used in src/arm64-ubuntu2204/qemu_files/get-disk-img.sh .

I tested this disk image by running gem5 simulations which booted up Ubuntu, then exited. In these simulations, the disk image was used as a DiskImageResource. I used the timing and atomic CPUs, and ran simulations on the +20GB and +3GB versions of the disk image.


ivanaamit:
We (UCD) will be working on Disk images for the next couple of weeks, and we'll get back to you when we have updates.

ivanaamit:
After following the steps you provided in the README.md, I have a couple of questions.

(1) Upon launching the disk image in step 5 and booting up Ubuntu, a login prompt appears. What should I enter as the username and password? 

(2) When attempting to execute Packer in step 6, I got the following error message: 
"-bash ./packer: cannot execute binary file: Exec format error."
I am currently using an x86 system. Was Packer meant to be run on an ARM host?

erin-le:
@ivanaamit 
For 1:  After running Packer and re-launching the QEMU instance, you should be automatically logged in as the root user. As such, the user doesn't need to enter a username and password.

For 2: Yes, the `packer-obtain.sh` included in this PR downloads the Linux ARM64 version of Packer. You can either use an ARM system or use the script attached below, which changes the download link and gets the Linux AMD version of Packer.
[packer-obtain.txt](https://github.com/gem5/gem5-resources/files/13048264/packer-obtain.txt)

I'll edit the README and/or some other files to make these things more clear.

codecrypt92:
@erin-le I cannot manage to get into the ssh access. Even after copying the  public keys and directing it to the right private key. I am stuck here. Any pointers to what I might be missing. I have given ~/.ssh/id_rsa 600 access, still no progress. 
On Packer terminal: 
ubuntu@localhost: Permission denied (publickey).
On QEMU boot: 
ubuntu login: ubuntu
Password:

Login incorrect
ubuntu login:
Login timed out after 60 seconds.

PR #9, opened 2023-09-20
resources: Add gfx902 targets to GPU applications

abmerop:
This adds gfx902 as a target for DNNMark's generate_cachefiles.py script and added the target to the Makefiles.  Note that the Makefiles can be overridden anyway with the HCC_AMDGPU_TARGET and this is the approach required to build HACC and PENNANT as the lines in the Makefile are actually ignored/do not exist. However, by adding the targets, resources can be built on real hardware with a simple 'make' command.

This is related to the discussion on the gem5 main repository about deprecating GCN3: https://github.com/orgs/gem5/discussions/306 . As a first step, we want to confirm that a Vega APU target works.  That target is gfx902.  I have tested everything in gem5-resources now and everything works as well as it does on gfx900 (dGPU target).  I have uploaded the scripts I used for testing here: https://github.com/abmerop/gem5/commit/89f7dba10062eeaf879e489ba76e87dc1baf8df9 . There is no intention of committing these scripts.  The commit is linked solely for reference.

### DIFF
--- a/src/gpu/DNNMark/generate_cachefiles.py
+++ b/src/gpu/DNNMark/generate_cachefiles.py
@@ -20,7 +20,7 @@ def parseArgs():
     parser.add_argument('--num-cus', default=4, type=int,
                         help='Number of CUs in simulated GPU')
     parser.add_argument('--gfx-version', default='gfx801',
-                        choices=['gfx801', 'gfx803', 'gfx900'],
+                        choices=['gfx801', 'gfx803', 'gfx900', 'gfx902'],
                         help='gfx version of simulated GPU')
 
     return parser.parse_args()
@@ -59,6 +59,8 @@ def insertFiles(con, options):
                   'gfx803': '-Wno-everything -Xclang '
                             '-target-feature -Xclang +code-object-v3',
                   'gfx900': '-Wno-everything -Xclang '
+                            '-target-feature -Xclang +code-object-v3',
+                  'gfx902': '-Wno-everything -Xclang '
                             '-target-feature -Xclang +code-object-v3'}
 
     with tempfile.TemporaryDirectory() as tmpdir:
--- a/src/gpu/heterosync/Makefile
+++ b/src/gpu/heterosync/Makefile
@@ -9,7 +9,7 @@ BIN_DIR := bin
 all: release
 
 release: $(SRC) | $(BIN_DIR)
-	$(HIP_PATH)/bin/hipcc -DGFX9 --amdgpu-target=gfx900 $(SRC) -o $(BIN_DIR)/$(EXECUTABLE)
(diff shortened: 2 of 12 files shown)


mattsinc (commented):
I'm surprised that MIOpen generates cachefiles properly for an APU (for DNNMark), given that it does not support APUs.  Your tests are finding that it generates something that works?

abmerop:
That's a good question.  I wanted to ask this:  How do you verify that DNNMark completes successfully?  Is there any output verification?  Right now it outputs nothing at all, but I can see from gem5 debug flags that it is actually running kernels.

mattsinc:
I guess there are a couple questions here:

1.  The only thing I've ever used to validate running was the debuginfo flag.  However, I don't know if this verifies the output or not (I suspect it doesn't).
2.  Are the kernels being generated working with APUs?  Or is it just generating gfx900-style kernels that have discrete GPU copies?
3. With gfx902, after generating, does DNNMark run to completion?

It sounds like the answer to 3 is yes.  With 2, I honestly don't know the best way to check this.

abmerop:
Hi Matt.  Yes the answer to 3 is yes.  For 2 I have been trying to run with dgpu and in theory there should be more tasks (memcpy + kernel + memcpy) but haven't had time yet. Just giving you an ACK that I read your comment from 2 days back.

mattsinc:
> Hi Matt. Yes the answer to 3 is yes. For 2 I have been trying to run with dgpu and in theory there should be more tasks (memcpy + kernel + memcpy) but haven't had time yet. Just giving you an ACK that I read your comment from 2 days back.

Sounds good.

abmerop:
Any chance we could get this moving along so I can turn on my tester and start catching bugs before 1 day before release? Since it is only modifying Makefiles the only "downside" is the binary is larger to hold the gfx902 ISA.

powerjg:
What needs to be done? What are we waiting on?

abmerop:
Still needs a review.  Not clear if this should be merged into develop or stable

powerjg:
Re: develop vs stable

If this will *only* work with gem5:develop, then it should be merged into develop. If this change works with gem5:stable, then it should be merged into stable. We would prefer merges into stable, but if this change depends on a change that's currently only on the develop branch of gem5 and it's not compatible with the stable branch, then we should merge it into gem5-resources:develop.

mattsinc (approved):

mattsinc:
> > Hi Matt. Yes the answer to 3 is yes. For 2 I have been trying to run with dgpu and in theory there should be more tasks (memcpy + kernel + memcpy) but haven't had time yet. Just giving you an ACK that I read your comment from 2 days back.
> 
> Sounds good.

It would be good if we answered this question, but I don't think is has to block this patch either.

PR #10, opened 2023-09-20
resources: SPEC 2017 benchmark suite on ARM64 Ubuntu 22.04 Disk Image

erin-le:
This pull request adds the files and instructions needed to make a ARM64 Ubuntu 22.04 disk image, and to install the SPEC 2017 benchmark suite on the disk image. The `qemu-files` directory contains the files for getting the disk image and launching it as a QEMU instance. The `packer-files` directory contains the files that Packer needs to build the disk image, as well as a script for obtaining Packer. README.md contains detailed instructions on how to build the disk image and install SPEC 2017 using the scripts and files provided.

I tested this by running a few gem5 simulations with the disk image that I made to figure out the process of building the disk image. Next, I launched the disk image as a QEMU instance and checked if the benchmarks that had built successfully would run. Finally, I built another disk image as part of writing the documentation, and used it as a resource to run another gem5 simulation.

### DIFF
--- a/src/spec-2017-ubuntu-22.04/README.md
+++ b/src/spec-2017-ubuntu-22.04/README.md
@@ -1,0 +1,199 @@
+# SPEC 2017 Benchmark Suite on ARM64 Ubuntu 22.04 Disk Image
+
+This file first documents how to build an ARM64 disk image with an installation of the SPEC 2017 benchmark suite, then describes how to use it as a gem5 resource.
+
+## Part 1: Making the disk image
+
+The following tutorial was followed, with some changes:
+
+https://github.com/takekoputa/hpc-disk-image/tree/main
+
+
+Note: All shell scripts should be run in the same directory that they are located in. That is, you should `cd` into the directory and run them with `./script_name.sh`. They were either provided by the tutorial as a file, or were commands provided in the tutorial that were made into shell scripts for ease of use.
+
+### Summary
+In short, you will need to:
+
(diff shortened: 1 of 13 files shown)


codecrypt92:
Hi @erin-le , 
The documentation you provided is not elaborate on adding the gem5-resource file. Do you have the launch script that you use to launch the disk with gem5 as a resource ?

powerjg:
`configs/example/gem5_library/x86-spec-cpu2017-benchmarks.py` should work. There's also a 2006 version as well.

codecrypt92:
@powerjg But that uses x86 boards and other configs that are used for x86. How is that going to work with ARM+KVM gem5  with SPEC2017 ?

powerjg:
Oh, sorry. I didn't look closely enough. We don't have any scripts out of the box for Arm+SPEC. However, you should be able to use the x86 one as a good example.

codecrypt92:
Hi Jason @powerjg , 
Thanks for the reply. However, I have tried multiple iterations of trying to generate an ARM version but it fails. Specifically, because I am trying to boot into a KVM version, and I am trying to follow arm-ubuntu-run-with-kvm.py, but it fails to boot. There are panic errors, segmentation faults. I am trying to start without a kvm, even that fails to boot. 
Are there any starting scripts or any efforts you know of to follow this for SPEC2017 (similar to x86-SPEC2017) @BobbyRBruce ?

PR #13, opened 2023-10-05
resources: Add OpenSBI Firmware as a Bootloader resource

hnpl:
This patch documents how to compile OpenSBI's `FW_JUMP` and use it as a bootloader in gem5.
The document also includes steps to build a linux kernel that works with gem5.

### DIFF
--- a/src/riscv-opensbi-alt/Dockerfile-linux
+++ b/src/riscv-opensbi-alt/Dockerfile-linux
@@ -1,0 +1,4 @@
+FROM --platform=$BUILDPLATFORM alpine:3.18.4
+
+# The Linux dependencies
+RUN apk add --no-cache build-base bc bison flex openssl paholelibssl perl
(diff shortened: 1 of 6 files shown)


hnpl:
@BobbyRBruce Let me know if we can reuse any of the cross compiler docker images in gem5 rather than having one here.

BobbyRBruce:
@hnpl : I played around with this (a bit too much) this morning. See my latest patch in the chain.

I'm not requesting you change your design but this is more to say, this is how I'd do it. It's much neater and, I think, provides more flexibility and stability.

Here we just need two things:

1. The Dockerfile define the build environment.
2. The docker-compose.yaml defining the platform/platforms targeted, and a clear, structured series of steps to create a resource.

I can't think of any build we'd want to do which doesn't work with this design. There's less bash scripts and manual effort. Assuming the user has docker setup correctly, `docker-compose up` is all they'd need to do.

BobbyRBruce (approved):

PR #5, opened 2023-08-15
resouces: add mmap support to FW and fix bugs

jebraun3:
resources: add mmap support to FW and bug fixes

PR #19, opened 2023-12-18
resources: Fix hipDeviceSynchronize calls

jebraun3:
Corrected hipDeviceSynchronize calls to prevent kernels from exiting prematurely

Change-Id: I068b6324f33d0de905477d03efa195e3068dc7c5

### DIFF
--- a/src/gpu/pannotia/bc/BC.cpp
+++ b/src/gpu/pannotia/bc/BC.cpp
@@ -206,11 +206,13 @@ int main(int argc, char **argv)
 
     // Initialization
     hipLaunchKernelGGL(HIP_KERNEL_NAME(clean_bc), dim3(grid), dim3(threads ), 0, 0, bc_d, num_nodes);
+    hipDeviceSynchronize();
 
     // Main computation loop
     for (int i = 0; i < num_nodes && i < MAX_ITERS; i++) {
         hipLaunchKernelGGL(HIP_KERNEL_NAME(clean_1d_array), dim3(grid), dim3(threads ), 0, 0, i, dist_d, sigma_d, rho_d,
                                             num_nodes);
+        hipDeviceSynchronize();
 
         // Depth of the traversal
         int dist = 0;
@@ -226,6 +228,7 @@ int main(int argc, char **argv)
 
             hipLaunchKernelGGL(HIP_KERNEL_NAME(bfs_kernel), dim3(grid), dim3(threads ), 0, 0, row_d, col_d, dist_d, rho_d, stop_d,
                                             num_nodes, num_edges, dist);
+            hipDeviceSynchronize();
 
             // Copy back the termination variable from the device
             hipMemcpy(&stop, stop_d, sizeof(int), hipMemcpyDeviceToHost);
@@ -235,22 +238,20 @@ int main(int argc, char **argv)
 
         } while (stop);
 
-        hipDeviceSynchronize();
-
         // Traverse back from the deepest part of the tree
         while (dist) {
             hipLaunchKernelGGL(HIP_KERNEL_NAME(backtrack_kernel), dim3(grid), dim3(threads ), 0, 0, row_trans_d, col_trans_d,
                                                 dist_d, rho_d, sigma_d,
                                                 num_nodes, num_edges, dist, i,
                                                 bc_d);
+            hipDeviceSynchronize();
 
             // Back one level
             dist--;
         }
-        hipDeviceSynchronize();
 	fprintf(stdout, "Completed iteration %d\n", i);
     }
-    hipDeviceSynchronize();
+
     //timer4 = gettime();
 
     // Copy back the results for the bc array
--- a/src/gpu/pannotia/color/coloring_max.cpp
+++ b/src/gpu/pannotia/color/coloring_max.cpp
@@ -230,10 +230,12 @@ int main(int argc, char **argv)
         hipLaunchKernelGGL(HIP_KERNEL_NAME(color1), dim3(grid), dim3(threads ), 0, 0, row_d, col_d, node_value_d, color_d,
                                      stop_d, max_d, graph_color, num_nodes,
                                      num_edges);
+        hipDeviceSynchronize();
 
         // Launch the color kernel 2
         hipLaunchKernelGGL(HIP_KERNEL_NAME(color2), dim3(grid), dim3(threads ), 0, 0, node_value_d, color_d, max_d, graph_color,
                                      num_nodes, num_edges);
+        hipDeviceSynchronize();
 
         err = hipMemcpy(&stop, stop_d, sizeof(int), hipMemcpyDeviceToHost);
         if (err != hipSuccess) {
@@ -244,7 +246,6 @@ int main(int argc, char **argv)
         graph_color++;
 
     }
-    hipDeviceSynchronize();
 
 //    double timer4 = gettime();
 
--- a/src/gpu/pannotia/color/coloring_maxmin.cpp
+++ b/src/gpu/pannotia/color/coloring_maxmin.cpp
@@ -215,6 +215,7 @@ int main(int argc, char **argv)
 
     // Initialize arrays
     hipLaunchKernelGGL(ini, dim3(grid), dim3(threads ), 0, 0, max_d, min_d, num_nodes);
+    hipDeviceSynchronize();
 
     // Main computation loop
     double timer3 = gettime();
@@ -233,10 +234,12 @@ int main(int argc, char **argv)
         hipLaunchKernelGGL(color1, dim3(grid), dim3(threads ), 0, 0, row_d, col_d, node_value_d, color_d,
                                      stop_d, max_d, min_d, graph_color,
                                      num_nodes, num_edges);
+        hipDeviceSynchronize();
 
         // Launch the color kernel 2
         hipLaunchKernelGGL(color2, dim3(grid), dim3(threads ), 0, 0, node_value_d, color_d, max_d, min_d,
                                      graph_color, num_nodes, num_edges);
+        hipDeviceSynchronize();
 
         err = hipMemcpy(&stop, stop_d, sizeof(int), hipMemcpyDeviceToHost);
         if (err != hipSuccess) {
@@ -247,7 +250,6 @@ int main(int argc, char **argv)
         graph_color = graph_color + 2;
 
     }
-    hipDeviceSynchronize();
 
     double timer4 = gettime();
 
--- a/src/gpu/pannotia/mis/mis.cpp
+++ b/src/gpu/pannotia/mis/mis.cpp
@@ -241,14 +241,17 @@ int main(int argc, char **argv)
         hipLaunchKernelGGL(HIP_KERNEL_NAME(mis1), dim3(grid), dim3(threads), 0, 0, row_d, col_d, node_value_d, s_array_d,
                                  c_array_d, min_array_d, stop_d, num_nodes,
                                  num_edges);
+        hipDeviceSynchronize();
 
         // Launch mis2
         hipLaunchKernelGGL(HIP_KERNEL_NAME(mis2), dim3(grid), dim3(threads), 0, 0, row_d, col_d, node_value_d, s_array_d,
                                  c_array_d, c_array_u_d, min_array_d, num_nodes,
                                  num_edges);
+        hipDeviceSynchronize();
 
         // Launch mis3
         hipLaunchKernelGGL(HIP_KERNEL_NAME(mis3), dim3(grid), dim3(threads), 0, 0, c_array_u_d, c_array_d, num_nodes);
+        hipDeviceSynchronize();
 
         // Copy the termination variable back
         err = hipMemcpy(&stop, stop_d, sizeof(int), hipMemcpyDeviceToHost);
@@ -259,8 +262,6 @@ int main(int argc, char **argv)
 
         iterations++;
     }
-
-    hipDeviceSynchronize();
 
     err = hipMemcpy(s_array, s_array_d, num_nodes * sizeof(int), hipMemcpyDeviceToHost);
     if (err != hipSuccess) {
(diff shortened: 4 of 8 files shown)


mattsinc (dismissed):

mattsinc:
@powerjg this would require creating a new binary.  Are you ok with merging in anyways?

powerjg:
We should probably make new binaries before it's merged. Otherwise, we could easily forget.

@Harshil2107 this is going to be on you :)

Harshil2107:
Should these binaries be built on top of the changes in #22?

mattsinc:
> Should these binaries be built on top of the changes in #22?



> Should these binaries be built on top of the changes in #22?

This is separate from #22 but if it's easier for you to update binaries once for both, that's fine with me

abmerop:
> Should these binaries be built on top of the changes in #22?

Just to clarify, #22 is only adding links to the prebuilts so as long as they are placed in the same link locations then no changes are required in #22.

abmerop (commented):
I guess I haven't reviewed yet...

What does "exiting prematurely" mean? Is gem5 exiting before the kernels finish?  Is the goal here to intentionally serialize the kernels?

mattsinc:
> I guess I haven't reviewed yet...
> 
> What does "exiting prematurely" mean? Is gem5 exiting before the kernels finish? Is the goal here to intentionally serialize the kernels?

@jebraun3 can chime in with more details, but my recollection is that we were seeing gem5 essentially treat the kernel launches as blocking, causing multiple kernels from the same stream to overlap and thus get incorrect data.  So we added the hipDeviceSynch's to prevent this.

Harshil2107:
Is this PR ready to be merged?

mattsinc:
> > I guess I haven't reviewed yet...
> > What does "exiting prematurely" mean? Is gem5 exiting before the kernels finish? Is the goal here to intentionally serialize the kernels?
> 
> @jebraun3 can chime in with more details, but my recollection is that we were seeing gem5 essentially treat the kernel launches as blocking, causing multiple kernels from the same stream to overlap and thus get incorrect data. So we added the hipDeviceSynch's to prevent this.

@jebraun3 reminder to reply here to @abmerop , before this can/should be merged in

ivanaamit:
@jebraun3, could you please let us know what is the status of this PR? Thank you.

mattsinc:
> > > I guess I haven't reviewed yet...
> > > What does "exiting prematurely" mean? Is gem5 exiting before the kernels finish? Is the goal here to intentionally serialize the kernels?
> > 
> > 
> > @jebraun3 can chime in with more details, but my recollection is that we were seeing gem5 essentially treat the kernel launches as blocking, causing multiple kernels from the same stream to overlap and thus get incorrect data. So we added the hipDeviceSynch's to prevent this.
> 
> @jebraun3 reminder to reply here to @abmerop , before this can/should be merged in

@abmerop do you want more details from @jebraun3 or is my explanation above sufficient?

abmerop:
My general concern is all of these syncs are going to tank the performance so I'm not sure how much I would trust this benchmark for any performance related data after this change.  But it seems to fix some problem I am not seeing, so I won't block it (if that is even possible on github)

mattsinc:
> My general concern is all of these syncs are going to tank the performance so I'm not sure how much I would trust this benchmark for any performance related data after this change. But it seems to fix some problem I am not seeing, so I won't block it (if that is even possible on github)

Why would it affect performance?  I guess for the CPU part?

abmerop:
> > My general concern is all of these syncs are going to tank the performance so I'm not sure how much I would trust this benchmark for any performance related data after this change. But it seems to fix some problem I am not seeing, so I won't block it (if that is even possible on github)
> 
> Why would it affect performance? I guess for the CPU part?

The sync is essentially a barrier which will prevent the new kernel dispatch packet from being placed in the AQL queue and therefore you would be paying an extra latency for sending the packet, ringing the doorbell, etc.  which would otherwise be done while the previous kernel was running

mattsinc:
> > > My general concern is all of these syncs are going to tank the performance so I'm not sure how much I would trust this benchmark for any performance related data after this change. But it seems to fix some problem I am not seeing, so I won't block it (if that is even possible on github)
> > 
> > 
> > Why would it affect performance? I guess for the CPU part?
> 
> The sync is essentially a barrier which will prevent the new kernel dispatch packet from being placed in the AQL queue and therefore you would be paying an extra latency for sending the packet, ringing the doorbell, etc. which would otherwise be done while the previous kernel was running

Right, so shaderActiveTicks wouldn't be affected (I think) but the sim_ticks would?

Regardless, @jebraun3 can you please add the details about what overlap you were seeing?  Ultimately it likely indicates something is wrong with the barrier bit, which in theory makes the device synch's not needed ...

jebraun3:
I ran into this first with the Floyd-Warshall benchmark where running FW in gem5 would produce different results from running it on the actual hardware. From what I remember the result array that was produced left the vast majority of the array untouched, so most of its values were still set to their initialized values. Adding the synchs to the benchmark got gem5 to produce the same result array as the actual hardware.

ivanaamit:
Should this be merged or not? If I don't hear back from you until the end of the day today, I will merge it.

mattsinc:
> Should this be merged or not? If I don't hear back from you until the end of the day today, I will merge it.

No, please do not merge until @jebraun3 , @abmerop , and myself come to a conclusion

abmerop:
I now think it's OK to merge this to get the correct result.

If I understand it correctly, the original code works gets correct result on hardware, new code gets correct result on hardware. But in gem5 the old code gets wrong result but new code gets correct result? 

If that is the case we should circle back and figure out why the original code is different.

mattsinc:
> I now think it's OK to merge this to get the correct result.
> 
> If I understand it correctly, the original code works gets correct result on hardware, new code gets correct result on hardware. But in gem5 the old code gets wrong result but new code gets correct result?
> 
> If that is the case we should circle back and figure out why the original code is different.

I believe this is correct -- on the real GPU the code works and gets the correct answer, in gem5 it gets the incorrect answer without this fix.  @jebraun3 please confirm.

I guess the question is: should we take this as a sign we need to investigate the barrier bit code again and try to fix it?  Or should we patch this even with the shortcomings you highlighted and investigate later?

abmerop:
> > I now think it's OK to merge this to get the correct result.
> > If I understand it correctly, the original code works gets correct result on hardware, new code gets correct result on hardware. But in gem5 the old code gets wrong result but new code gets correct result?
> > If that is the case we should circle back and figure out why the original code is different.
> 
> I believe this is correct -- on the real GPU the code works and gets the correct answer, in gem5 it gets the incorrect answer without this fix. @jebraun3 please confirm.
> 
> I guess the question is: should we take this as a sign we need to investigate the barrier bit code again and try to fix it? Or should we patch this even with the shortcomings you highlighted and investigate later?

A barrier problem was my first thought.  Specifically of the AND variety.  And yes I think we should look into that.

mattsinc:
> > > I now think it's OK to merge this to get the correct result.
> > > If I understand it correctly, the original code works gets correct result on hardware, new code gets correct result on hardware. But in gem5 the old code gets wrong result but new code gets correct result?
> > > If that is the case we should circle back and figure out why the original code is different.
> > 
> > 
> > I believe this is correct -- on the real GPU the code works and gets the correct answer, in gem5 it gets the incorrect answer without this fix. @jebraun3 please confirm.
> > I guess the question is: should we take this as a sign we need to investigate the barrier bit code again and try to fix it? Or should we patch this even with the shortcomings you highlighted and investigate later?
> 
> A barrier problem was my first thought. Specifically of the AND variety. And yes I think we should look into that.

Right, the question is: should we merge this and then fix (and revert later), or just not merge this and try to fix?  My concern without merging is if people are using it now, they are not getting the correct answer.  And it's unknown how long a fix would take (maybe long, maybe short).

abmerop:
I did not see any problem when running in FS mode, so maybe this is only happening in SE mode?  @jebraun3  what does the error look like?  The application doesn't seem to print anything

jebraun3:
What @mattsinc said is correct, on the real GPU the code works and gets the correct answer but in gem5 it gets the incorrect answer. An error is only printed when FW is run with error checking, otherwise there is no indication of the error. When printed, the error should be a list of the locations of the mismatches followed by "WARNING: Produced incorrect results!".

mattsinc:
> What @mattsinc said is correct, on the real GPU the code works and gets the correct answer but in gem5 it gets the incorrect answer. An error is only printed when FW is run with error checking, otherwise there is no indication of the error. When printed, the error should be a list of the locations of the mismatches followed by "WARNING: Produced incorrect results!".

@jebraun3 It looks like FW's change is not in this commit, is that intentional?  I was looking because I believe FW requires the user to manually change the output to check it, right?

ivanaamit:
Hi, what is the status of this PR?

mattsinc:
> Hi, what is the status of this PR?

Please just leave it open, I need to talk with @jebraun3 and @abmerop about it

abmerop:
This is fairly ancient at this point, but looking into it, it seems the HIP runtime (well rocCLR) is smart enough to know if `hipDeviceSynchronize` is necessary or not for a given device.

Therefore my previous comment no longer applies and I am fine with this PR.

BobbyRBruce:
When this is confirmed "good to go" can you please ping @Harshil2107 to rebuild these binaries and upload them? I want to make sure the source and the binaries are in-sync here.

mattsinc:
> When this is confirmed "good to go" can you please ping @Harshil2107 to rebuild these binaries and upload them? I want to make sure the source and the binaries are in-sync here.

@jebraun3 is away on internship this semester and I'm guessing will not be able to reply too much.  With the context @abmerop provided (thanks!) I think this patch is ok as is to go now, @BobbyRBruce , unless we want to switch it to be on stable instead of develop?

@jebraun3 it looks like we'll need a similar patch with the FW updates, but since you are away I'd rather we integrate this then wait more time for that one benchmark.

BobbyRBruce:
Anything on the stable branch needs to be compatible with the current version of gem5 (that is, gem5 v24.0). develop is used to submit changes that should only be merged into stable upon the next release of gem5 (i.e., it requires some feature in the upcoming release and would fail on 24.0).

In short, go for stable if the resource works with 24.0.

mattsinc:
> Anything on the stable branch needs to be compatible with the current version of gem5 (that is, gem5 v24.0). develop is used to submit changes that should only be merged into stable upon the next release of gem5 (i.e., it requires some feature in the upcoming release and would fail on 24.0).
> 
> In short, go for stable if the resource works with 24.0.

This works with both stable and develop.

BobbyRBruce:
If this change works with v24.0 then I'd recommend putting it on stable but if you keep on the develop branch the only downside is you'll have to wait month until the next release when the gem5-resources develop branch is merged into the stable branch.

It's not worth worrying about either way.

mattsinc:
> If this change works with v24.0 then I'd recommend putting it on stable but if you keep on the develop branch the only downside is you'll have to wait month until the next release when the gem5-resources develop branch is merged into the stable branch.
> 
> It's not worth worrying about either way.

Since @jebraun3 is away on internship and thus I don't think he'll be able to move it soon, unless there is an easy way to switch branches on Github here, we should just merge on develop.  Is there such an easy way?

jebraun3:
I updated the PR to merge into stable. 

> @jebraun3 it looks like we'll need a similar patch with the FW updates, but since you are away I'd rather we integrate this then wait more time for that one benchmark.

It looks like FW already has this update from #5.

Harshil2107:
Is this PR ready and just waiting on resources to be created?

If so, I noticed there have been changes to some binaries in Pannotia, such as color. Should this PR be merged into develop first, and then the new binaries generated?

mattsinc:
> Is this PR ready and just waiting on resources to be created?
> 
> If so, I noticed there have been changes to some binaries in Pannotia, such as color. Should this PR be merged into develop first, and then the new binaries generated?

Yes, that all sounds right.

Harshil2107:
I updated the target branch to develop, if every thing looks good I can remake the binaries.

mattsinc:
> I updated the target branch to develop, if every thing looks good I can remake the binaries.

Yes, please go ahead, thanks!

Harshil2107:
I remade the updated binaries and uploaded them to the bucket, here are the links:

https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/bc.gem5_20250514
https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/color_maxmin.gem5_20250514
https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/color_max.gem5_20250514
https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/pagerank.gem5_20250514
https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/mis_hip.gem5_20250514
https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/pagerank_spmv.gem5_20250514
https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/sssp.gem5_20250514
https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/sssp_ell.gem5_20250514

mattsinc:
> I remade the updated binaries and uploaded them to the bucket, here are the links:
> 
> https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/bc.gem5_20250514 https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/color_maxmin.gem5_20250514 https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/color_max.gem5_20250514 https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/pagerank.gem5_20250514 https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/mis_hip.gem5_20250514 https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/pagerank_spmv.gem5_20250514 https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/sssp.gem5_20250514 https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/pannotia/sssp_ell.gem5_20250514

thanks!

mattsinc:
@BobbyRBruce it looks like you need to re-review this one as it's blocked by your previously requested changes.

BobbyRBruce:
@Harshil2107 why is this going to the develop branch?

Are these changes not applicable to stable?

Harshil2107:
There are changes made to color binaries in develop on top of which we want these changes.

BobbyRBruce:
Ok, thanks for clarifying

BobbyRBruce (approved):

mattsinc:
thanks all!

PR #23, opened 2024-02-18
resources: Add documentation for Getting Started Suite

kunpai:
This patch adds the make.def files for NPB for different ISAs, and adds instructions on compiling (and cross-compiling) the LLVM Test Suite.
This is the process that was used to compile the binaries used in the "Getting Started" Suite as well.

### DIFF
--- a/src/getting-started/BUILDING.md
+++ b/src/getting-started/BUILDING.md
@@ -1,0 +1,270 @@
+# Building the Getting Started Suite
+
+The Getting Started Suite is designed to provide a set of pre-compiled familiar workloads to help one get started with gem5.
+
+This document will guide you through the process of building the binaries for the Getting Started Suite.
+You may want to follow this guide if you want to customize these binaries, use a different compiler, or compile them to a new ISA.
+Note that as described in the README.md file there are precompiled binaries available from the gem5 resources binary repository.```
+
+## Table of Contents
+
+- [Building the Suite Binaries](#building-the-suite-binaries)
+  - [gem5-resources](#gem5-resources)
+  - [GAPBS](#gapbs)
+  - [LLVM Test Suite](#llvm-test-suite)
+    - [Cross-compiling the LLVM test suite](#cross-compiling-the-llvm-test-suite)
+  - [NAS Parallel Benchmarks](#nas-parallel-benchmarks)
+    - [Using the Suite Binaries](#using-the-suite-binaries)
+
+## Building the Suite Binaries
+
+### gem5-resources
+
+First, clone the gem5-resources repository:
+
+```bash
+git clone https://github.com/gem5/gem5-resources.git
+```
+
+Next, checkout the following commit:
+
+```bash
+cd gem5-resources
+git checkout dc3e57f6de56496188b06404c7bd4dabee8fffef
+```
+
+Next, navigate to the `src/matrix-multiply` directory:
+
+```bash
+cd gem5-resources/src/matrix-multiply
+```
+
+Then, build the matrix multiply binary:
+
+```bash
+make
+```
+
+The x86 version of the matrix multiply binary was compiled using the `gcc` compiler native to the system.
+The version of `gcc` used to compile the matrix multiply binary was `gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0`.
+
+The ARM version of the matrix multiply binary was compiled using the `gcc` compiler native to the system.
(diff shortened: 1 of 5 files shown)


powerjg:
This is good documentation, but I don't think it goes far enough.

I think you should add a new directory for the "getting started" suite that you made. Then, the documentation in that directory should be detailed enough for *anyone else in the world* to reproduce *the exact binaries that you have.* Basically, you need to write it as though you're giving directions to a computer.

I suggest creating a `getting_started/BUILDING.md` file and a `getting_started/README.md` file. The `BUILDING.md` file should contain all of the details need to build the workloads in the suite and the `README.md` should explain the point of the suite.

When writing this documentation, try to put yourself in someone else's shoes. E.g., pretend you are a first year grad student and you understand what "compiling" is and you understand what "gem5" is, but you don't know what the "getting started suite" is or how to build it, use it, and (if possible) modify it.

src/getting-started/BUILDING.md
```diff
@@ -0,0 +1,186 @@
...
+Next, run make to build the BFS and TC binaries:
+
+```bash
+make
+```
+
+If you want to build the binaries for a specific architecture, you need to cross-compile the binaries.
+For example, to build the binaries for RISC-V, you need to edit the `Makefile` and add a variable called `CXX` and set it to the RISC-V cross-compiler.
+An example of the `CXX` variable is `CXX=riscv64-linux-gnu-g++`.
+Then, run the `make` command.
```
powerjg: This is OK, but it's not reproducible. I would like you all to be thinking about how to make these resources *reproducible* by other people (or, just as importantly, by you in 2 years).

src/getting-started/BUILDING.md
```diff
@@ -0,0 +1,186 @@
...
+
+### LLVM Test Suite
+
+On a native system, the LLVM test suite can be compiled using the following:
```
powerjg, line 118: Where do you get the code?

src/getting-started/BUILDING.md
```diff
@@ -0,0 +1,186 @@
...
+```bash
+cmake --build build --target <test_name>
+```
+
+where `<test_name>` is the name of the test that you want to build.
+
+For example, to build the `MultiSource/Applications/minisat` test, use the following command:
+
+```bash
+cmake --build build --target minisat
```
powerjg, line 143: What workloads are you building from this suite?

src/getting-started/BUILDING.md
```diff
@@ -0,0 +1,186 @@
...
+
+```bash
+cmake --build build --target <test_name>
+```
+
+where `<test_name>` is the name of the test that you want to build.
+
+### NAS Parallel Benchmarks
+
+A pre-requisite for this tutorial is to have the NAS Parallel Benchmarks installed.
```
powerjg, line 184: Why are we not using the code in this repo?
https://github.com/gem5/gem5-resources/tree/stable/src/npb/disk-image/npb/npb-hooks/NPB3.3.1/NPB3.3-SER

src/getting-started/BUILDING.md
```diff
@@ -0,0 +1,194 @@
+# Building the Getting Started Suite
+
+The Getting Started Suite is designed to provide a set of pre-compiled familiar workloads to help one get started with gem5.
+
+This document will guide you through the process of building the binaries for the Getting Started Suite.
```
powerjg, line 5: We should make it clear that these directions are not needed for everyone. Maybe something like

```suggestion
This document will guide you through the process of building the binaries for the Getting Started Suite.
You may want to follow this guide if you want to customize these binaries, use a different compiler, or compile them to a new ISA.
Note that as described in the README.md file there are precompiled binaries available from the gem5 resources binary repository.```

File: src/getting-started/BUILDING.md [hunk 6]
```diff
@@ -0,0 +1,194 @@
...
+
+## Building the Suite Binaries
```
powerjg, line 19: Some navigational information would be good (e.g., table of contents). 

Also, there is not enough information here to actually reproduce the binaries that you created. We don't control some of these repos, so the code could change. Also, you haven't specified what compilers were used to create the binaries or where you got the compilers.

It's not going to be possible to describe everything in enough detail to make the build fully reproducible (https://reproducible-builds.org/docs/definition/), but we could be closer.

powerjg (approved):

Harshil2107 (approved):

PR #26, opened 2024-03-22
resources: Remove deleted GPUFS files from develop

abmerop:
These were already removed from stable. The disk image and README are wildly out of date. There is already a disk image folder but the README needs to be rewritten.


mattsinc (approved):
we should really make sure to replace them with something, at least making it even with stable?

abmerop:
> we should really make sure to replace them with something, at least making it even with stable?

Yes.  I did not realize the README is also on stable so that will have to be removed too.  In terms of replacement, my opinion is the documentation for general GPUFS should go on the website, since there is nothing left in this folder besides the Vega10 MMIO trace which is still needed for now (I would like to officially move to MI200 though).

mattsinc:
> > we should really make sure to replace them with something, at least making it even with stable?
> 
> Yes. I did not realize the README is also on stable so that will have to be removed too. In terms of replacement, my opinion is the documentation for general GPUFS should go on the website, since there is nothing left in this folder besides the Vega10 MMIO trace which is still needed for now (I would like to officially move to MI200 though).

I'm not sure a document on the website is any better or worse than a README, but as long as it's up-to-date I guess that's all that matters.

PR #29, opened 2024-04-13
resources: Update GPUFS disk to ROCm 6.1, HCL format

abmerop:
Update the disk image creation scripts to use the HCL packer formatted, modified from the x86-ubuntu gem5 resource. The major feature of this is a simplified one-step command to build the disk image.

In addition the GPU package versions are bumped to ROCm 6.1, PyTorch 2.2.2, and TensorFlow 2.14.0

The previous ROCm 5.4.2 resource is removed and the new disk image uses an unversioned x86-ubuntu-gpu-ml directory.


abmerop:
Not sure.  Do I need to update resources.json for this?

src/x86-ubuntu-gpu-ml/scripts/rocm-install.sh
```diff
@@ -0,0 +1,86 @@
...
+# OS: Linux
+# Package: Pip
+# Language: Python
+# Compute Platfrom: ROCm 5.7 (Note: No ROCm 6.0 available)
+pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/rocm5.7
+
+# See https://pypi.org/project/tensorflow-rocm/#description
+# Datasets are also installed as gem5 has no internet connection.
+pip install tensorflow-rocm
+pip install tensorflow_datasets
```
powerjg: Is it possible to include the specific versions here? I worry a bit about someone re-running this in a few weeks and generating a significantly different image. 

It's ok if the image isn't perfectly reproducible, but if it's possible to make it less likely to produce a different (and incompatible) image, that would be greatly appreciated!

src/x86-ubuntu-gpu-ml/x86-ubuntu-gpu-ml.pkr.hcl
```diff
@@ -0,0 +1,84 @@
...
+  provisioner "file" {
+    destination = "/root/roms/"
+    source      = "files/mi200.rom"
+  }
+
+  provisioner "file" {
+    source      = "/home/gem5/vmlinux-gpu-ml"
+    destination = "vmlinux-gpu-ml"
+    direction   = "download"
+  }
```
powerjg, line 83: @Harshil2107 we should do something like this in the base image to grab the kernel from the disk. I can give you more details.

src/x86-ubuntu-gpu-ml/files/run_gem5_app.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
powerjg, line 1: Any reason this is totally different from the base image's script? It's OK if it needs to be different, but if there's no need, then I would prefer the execution path of each disk to be as similar as possible.

src/x86-ubuntu-gpu-ml/README.md
```diff
@@ -0,0 +1,77 @@
...
+## Disk Image with QEMU
+
+The disk image is setup to automatically log in by default in gem5.
+If you are using an emulator such as QEMU to work with the disk, the login information is:
+
+- username: gem5
+- password: 12345
+
+## Example gem5 commands
+
```
powerjg, line 37: Can you add a description of the `m5 exit` calls and how users should expect the disk to behave?

I would prefer, though it's not required, to have all of our disks match as closely as possible. E.g., on the new `ubuntu-22.04` disk we have an exit after kernel boot, after systemd, etc.

src/x86-ubuntu-gpu-ml/BUILDING.md
```diff
@@ -0,0 +1,63 @@
...
+
+You can mount the disk image to see what is inside or to add additional files.
+Use the following command to mount the disk image:
+
+```sh
+mkdir mount
+sudo mount -o loop,offset=1048576 disk-image/x86-ubuntu-gpu-ml mount
+```
+
+This is particularly useful to copy ML datasets to the disk image.
```
powerjg: Can you say instead that this is useful for testing changes? I think we should encourage people to make changes in the `rocm-install.sh` file instead of manually changing the disk so that their changes are reproducible.

I don't know about you, but my experience has been that when I make manual changes to disks like this, I get to the "paper publishing time" and I have *no idea* what's on the disk and what isn't on the disk. I've gotten into some embarrassing situations where I can't reproduce plots in my papers because I didn't track what was on the disk carefully.

powerjg (commented):
Looks great. I believe everything below are just suggestions, so feel free to let me know what you don't want to do anything.

mattsinc (commented):
I don't have any major issues with this, but since @powerjg has some questions, I'll wait on approving.

I'll also need to test this locally, but that is unlikely to happen this week.

src/x86-ubuntu-gpu-ml/x86-ubuntu-gpu-ml.pkr.hcl
Harshil2107: I assume, we also want to get the kernel as a separate output and make a workload i.e. a boot workload using the diskimage + the extracted kernel?

Harshil2107:
> Not sure. Do I need to update resources.json for this?

If we need to make a brand new diskimage, then I can bump the version of the existing one in resources (MongoDB) or make one if it doesn't exist.

src/x86-ubuntu-gpu-ml/scripts/rocm-install.sh
abmerop: Good suggestion!  I forgot about this.  Made the change. tensorflow_datasets changed to tensorflow-datasets, but I rebuilt the image to confirm it still works.

src/x86-ubuntu-gpu-ml/BUILDING.md
abmerop: :)  ... familiar.  I changed the wording around this

src/x86-ubuntu-gpu-ml/README.md
abmerop: `m5 exit`  ends the simulation.  Anything else would require changes to the config scripts. I wasn't aware that this was changing.  Any reason not to create a new m5op?

src/x86-ubuntu-gpu-ml/files/run_gem5_app.sh
abmerop: I didn't really follow that script.  We want to read a script, run, and exit.  If it fails (e.g., QEMU) this will simply drop to shell.  I'm not sure why all the other stuff is needed.  To get interactive for example, we just run `bash --norc` as the application (which avoids an infinite loop reading the bashrc, which calls m5 readfile, etc.)

One other major difference from the base image I found after a lot of testing is that we must login a root user for GPU. This is because we need to modprobe the driver.  We have that blacklisted when Linux boots.  The reason for that is we need to workaround KVM not having memory holes and we copy the GPU bios to the x86 VGA ROM region in gem5 before modprobe.  The GPU driver supports multiple ways to read the GPU bios, but some of those are not compatible with some upcoming products we'd like to support in gem5

powerjg: Hmm, I would like to understand more of what you're saying here. If there's a better way to accomplish our goals, I would love to hear it.

We're trying to avoid logging in as root because it screws up other things (e.g., MPI). Would it work to make sudo passwordless and use `sudo modprobe`? I think that could be something that we change in the base image :)

Edit: No need to change this. But if `sudo modprobe` would work for you, let's make that change later. @Harshil2107 note this down.

src/x86-ubuntu-gpu-ml/README.md
powerjg: Yeah, we're working on creating new m5ops for each of these different points in the process. @BobbyRBruce is assigned this. For now, let's just document what we're doing and then Bobby or Harshil can fixup this disk image once we've finalized the new ops.

src/x86-ubuntu-gpu-ml/x86-ubuntu-gpu-ml.pkr.hcl
powerjg: Yeah, right :). I'm going to resolve this comment.

powerjg (approved):

Harshil2107 (dismissed):
lgtm

powerjg (dismissed):

abmerop:
> > Not sure. Do I need to update resources.json for this?
> 
> If we need to make a brand new diskimage, then I can bump the version of the existing one in resources (MongoDB) or make one if it doesn't exist.

We could reuse the "x86-gpu-fs-img" resource... Although maybe it makes sense to rename to the same name as the directory

abmerop:
Thanks for the approvals, but ROCm 6.1 was released yesterday so I am going to iterate on this one more time

abmerop:
Bumped to ROcm 6.1 and added cmake package which helps with testing

abmerop:
Bump

Harshil2107 (approved):
lgtm

PR #30, opened 2024-04-17
add riscv ubuntu 22.04 image

Harshil2107:
This diskimage follows similar structure as the new x86 image. This image does not have readme.md or building.md as this will be merged with the x86 and arm images to a single source.


src/riscv-ubuntu-22.04/riscv-ubuntu.pkr.hcl
```diff
@@ -0,0 +1,89 @@
...
+
+source "qemu" "initialize" {
+  cpus             = "4"
+  disk_size        = "5000"
+  format           = "raw"
+  headless         = "true"
+  disk_image       = "true"
+  http_directory   = "http"
+  boot_command = [
+                  "<wait90>",
```
powerjg: Where does this 90 come from? Is there any way to know when the it's ready?

src/riscv-ubuntu-22.04/riscv-ubuntu.pkr.hcl
```diff
@@ -0,0 +1,89 @@
...
+                  "ubuntu<enter><wait>",
+                  "ubuntu<enter><wait>",
+                  "ubuntu<enter><wait>",
+                  "12345678<enter><wait>",
+                  "12345678<enter><wait>",
+                  "<wait>",
+                ]
+  iso_checksum     = "sha256:0363f6fceef18dfc106379a7487af6648dfc7540423d57a36080d9fc53633dcf"
+  iso_urls         = ["/home/harshilp/forks/gem5-resources-repo/fresh-riscv-img.img"]
```
powerjg: This can't be committed with you local file paths

src/riscv-ubuntu-22.04/riscv-ubuntu.pkr.hcl
```diff
@@ -0,0 +1,89 @@
...
+}
+
+variable "ssh_username" {
+  type    = string
+  default = "ubuntu"
+}
+
+source "qemu" "initialize" {
+  cpus             = "4"
+  disk_size        = "5000"
```
powerjg, line 27: Do we have to use a disk this big?

powerjg (commented):
Please put newlines at the end of files.

Check to see if any of my other comments on the Arm version need any updates here.

src/riscv-ubuntu-22.04/riscv-ubuntu.pkr.hcl
Harshil2107: It took around 60 seconds to boot so I added a 30 second buffer. The only way to know when its booted is to have the display.

src/riscv-ubuntu-22.04/riscv-ubuntu.pkr.hcl
Harshil2107: About 4.1GB is used so I have 900MB for extra things that users might want to install.

src/riscv-ubuntu-22.04/riscv-ubuntu.pkr.hcl
powerjg: You should document this in the README. You should also document what happens if someone else's computer is slower than yours.

src/riscv-ubuntu-22.04/.gitignore
```diff
@@ -0,0 +1,4 @@
+**packer
+**packer_*.zip
+packer_cache/
+disk-image/
```
powerjg: Missing newline. Please look through your diffs and do a self code review when pushing commits.

src/riscv-ubuntu-22.04/riscv-ubuntu.pkr.hcl
Harshil2107: I updated it to be a relative path from the current directory and the `build.sh` file downloads the image.

src/riscv-ubuntu-22.04/build.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Copyright (c) 2024 The Regents of the University of California.
+# SPDX-License-Identifier: BSD 3-Clause
+
+PACKER_VERSION="1.10.0"
+
+if [ ! -f ./packer ]; then
+    wget https://releases.hashicorp.com/packer/${PACKER_VERSION}/packer_${PACKER_VERSION}_linux_amd64.zip;
+    unzip packer_${PACKER_VERSION}_linux_amd64.zip;
+    rm packer_${PACKER_VERSION}_linux_amd64.zip;
```
powerjg, line 11: This should not be addressed in the current changeset, but it's now becoming a very high priority to address. We can't have the host CPU (i.e., which packer to download) tied to the guest that we are working on. Please open an issue about this. I can explain in more detail, if needed.

src/riscv-ubuntu-22.04/scripts/post-installation.sh
```diff
@@ -0,0 +1,95 @@
...
+rm -rf gem5
+echo "Done building and installing gem5-bridge (m5) and libm5"
+
+# You can extend this script to install your own packages here.
+
+# Disable network by default
+echo "Disabling network by default"
+echo "See README.md for instructions on how to enable network"
+mv /etc/netplan/50-cloud-init.yaml /etc/netplan/50-cloud-init.yaml.bak
+# netplan apply
```
powerjg: Why is this commented out?

src/riscv-ubuntu-22.04/riscv-ubuntu.pkr.hcl
```diff
@@ -0,0 +1,91 @@
...
+  format           = "raw"
+  headless         = "true"
+  disk_image       = "true"
+  http_directory   = "http"
+  boot_command = [
+                  "<wait120>",
```
powerjg, line 33: This should be documented in the BUILDING.md

powerjg (changes requested):
This looks OK. However, it's missing a README.md and a BUILDING.md. Also, IIRC, there are changes to files that were requested in other PRs (e.g., comments on `popd` lines) that have not been reflected here. (The newlines are fine, btw)

src/riscv-ubuntu-22.04/scripts/post-installation.sh
Harshil2107: for 24.04 it causes the build process to hang as when this line is reached and the change to the yamls is applied instantly and it cuts off the ssh connection causing the build to fail. By disabling the networkd wait online service , we no longer need to netplan apply immidiately, so I commented the line. I will remove the line as it is not necessary.

src/riscv-ubuntu-22.04/riscv-ubuntu.pkr.hcl
Harshil2107: I have added the BUILDING.md file, this follows the same structure as the x86 BUILDING.md but with modifications pertaining to the riscv image.

src/riscv-ubuntu-22.04/build.sh
Harshil2107: I will make an issue. Just to clarify, the issue is that we are replying on the host system to figure out which packer to download but there should be some automated way? or some sort of error message when wrong packer is downloaded?

powerjg: Either. You could certainly do the former.

In fact, we could just do `apt install packer` and not have this download thing going on...

powerjg (dismissed):
Thanks for the updates!

src/riscv-ubuntu-22.04/scripts/post-installation.sh
```diff
@@ -0,0 +1,95 @@
...
+cp /etc/fstab /etc/fstab.backup
+
+# Ensure the root filesystem does not remount as read-only on errors
+sed -i 's/\(LABEL=cloudimg-rootfs.*\)errors=remount-ro\(.*\)/\1 \2/' /etc/fstab
+
+# Comment out the mounting of the /boot/efi partition
+sed -i '/\/boot\/efi/s/^/#/' /etc/fstab
+
+# Optional: Display /etc/fstab to verify changes
+cat /etc/fstab
```
powerjg: Just for the public record, I would feel more comfortable if we didn't make these changes to `fstab`. I'm particularly concerned about the change to remount in `rw` mode even if there are errors.

src/riscv-ubuntu-22.04/build.sh
Harshil2107: https://github.com/gem5/gem5/issues/1252

powerjg (approved):

PR #31, opened 2024-04-18
resources: ubuntu-x86 move installing packages to post install

Harshil2107:

PR #35, opened 2024-05-15
resources: Move GPUFS hcl packages, PyTorch 2.3.0

abmerop:
Move the packages from http/user-data to the install script similar to
(https://github.com/gem5/gem5-resources/pull/31). Bump PyTorch to version 2.3.0. This is the first version to
support MI300. An MI300 config should be released with v24.0.


abmerop:
If the resource hasn't been made yet / not too much work to do again, we should get this in as we can run PyTorch on MI300 configs which would be a big step in the release

mattsinc (approved):

PR #36, opened 2024-05-28
gem5-resources: ARM Ubuntu 24.04 disk image

ivanaamit:
- [x] Tested 
- [x] Uploaded to the cloud bucket
- [ ] Added to gem5 resources website.

### DIFF
--- a/src/arm-ubuntu/arm-ubuntu-24-04.pkr.hcl
+++ b/src/arm-ubuntu/arm-ubuntu-24-04.pkr.hcl
@@ -1,0 +1,95 @@
+packer {
+  required_plugins {
+    qemu = {
+      source  = "github.com/hashicorp/qemu"
+      version = "~> 1"
+    }
+  }
+}
+
+variable "image_name" {
+  type    = string
+  default = "arm-ubuntu"
+}
+
+variable "ssh_password" {
+  type    = string
+  default = "12345"
+}
+
+variable "ssh_username" {
+  type    = string
+  default = "gem5"
+}
+
+source "qemu" "initialize" {
+  boot_command     = [
+                      "c<wait>",
+                      "linux /casper/vmlinuz autoinstall ds=nocloud-net\\;s=http://{{.HTTPIP}}:{{.HTTPPort}}/ --- ",
+                      "<enter><wait>",
+                      "initrd /casper/initrd",
+                      "<enter><wait>",
+                      "boot",
+                      "<enter>",
+                      "<wait>"
+                      ]
+  cpus             = "4"
+  disk_size        = "4600"
+  format           = "raw"
+  headless         = "true"
+  http_directory   = "http-24-04"
+  iso_checksum     = "sha256:d2d9986ada3864666e36a57634dfc97d17ad921fa44c56eeaca801e7dab08ad7"
+  iso_urls         = ["https://cdimage.ubuntu.com/releases/24.04/release/ubuntu-24.04-live-server-arm64.iso"]
+  memory           = "8192"
+  output_directory = "disk-image-24-04"
+  qemu_binary      = "/usr/bin/qemu-system-aarch64"
+  qemuargs         = [  ["-boot", "order=dc"],
+                        ["-bios", "./files/flash0.img"],
+                        ["-cpu", "host"],
+                        ["-enable-kvm"],
+                        ["-machine", "virt"],
+                        ["-machine", "gic-version=3"],
+                        ["-vga", "virtio"],
+                        ["-device","virtio-gpu-pci"],
+                        ["-device", "qemu-xhci"],
+                        ["-device","usb-kbd"],
+
+                      ]
+  shutdown_command = "echo '${var.ssh_password}'|sudo -S shutdown -P now"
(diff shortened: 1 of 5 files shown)


src/arm-ubuntu/arm-ubuntu-24-04.pkr.hcl
```diff
@@ -0,0 +1,95 @@
...
+                        ["-cpu", "host"],
+                        ["-enable-kvm"],
+                        ["-machine", "virt"],
+                        ["-machine", "gic-version=3"],
+                        ["-vga", "virtio"],
+                        ["-device","virtio-gpu-pci"],
+                        ["-device", "qemu-xhci"],
+                        ["-device","usb-kbd"],
+
+                      ]
```
powerjg, line 57: This assumes you're running on arm hardware. Is this the same as the 22.04 image? @Harshil2107 

We need to at a minimum document, but it would be better to provide an automatic approach so that you can build this image on an Arm machine or x86 machine.

src/arm-ubuntu/files/gem5_init.sh
```diff
@@ -19,7 +19,6 @@ fi
 if [[ $no_systemd == false ]]; then
     # gem5-bridge exit signifying that systemd will be booted
     printf "Starting systemd...\n"
-    gem5-bridge exit
```
powerjg, line 22: Why did you remove this line?

src/arm-ubuntu/arm-ubuntu-24-04.pkr.hcl
Harshil2107: yes, This is the same as the 22.04 image. Should I add an equivalent qemu command to documentation? Packer does give the qemu command it uses when we run the build with `PACKER_LOG=INFO`, and I have tried using the same args and I can boot the image in qemu.

Harshil2107: I can figure out the args needed to build on an x86 machine and we can add it as a comment or as another packer file to build without kvm?

powerjg: Maybe we should have two packerfiles (or if there's a way to share things between files, even better!) one for kvm and one for "other" host.

src/arm-ubuntu/files/gem5_init.sh
ivanaamit: In Harshil's updated README, there is no exit event at that point.

src/arm-ubuntu/arm-ubuntu-24-04.pkr.hcl
```diff
@@ -0,0 +1,95 @@
+packer {
```
powerjg, line 1: I don't know if this is in other files or not, but if it's possible we should have a copyright block.

src/arm-ubuntu/arm-ubuntu-24-04.pkr.hcl
powerjg: So this is getting solved in another PR?

src/arm-ubuntu/files/gem5_init.sh
powerjg: @Harshil2107 can you confirm that all of the other init scripts no longer have this exit?

src/arm-ubuntu/arm-ubuntu-24-04.pkr.hcl
Harshil2107: I will look into it, if I can add comments (should be able to), then I will add the copyright statement in the refactor PR, as x86 packer files also doesn't have a copyright statement.

Harshil2107:
For the other 2 comments:

- yes, I will add the qemu args for no kvm in the refactor PR when I merge the arm packer files 
- yes, this is an extra exit event that we had decided we don't want, none of the other disk images have this exit event.

powerjg (approved):

ncde:
Hello, I try to follow the README to create the arm-ubuntu image myself to install custom sw in qemu to use in gem5 later.
I could not find in the repository for example arm-ubuntu.json file, where is this located?
Also there are some unused files in gem5-resources-stable/src/arm/ubuntu/files such as after_boot.sh, when is this used?

do you have an updated README file for the instructions?

Sorry I write comment to your commit but I could not find the place to create an issue probably it is disabled.
Best,

Harshil2107:
Hi,

The source to build the arm image is at https://github.com/gem5/gem5-resources/tree/stable/src/arm-ubuntu

We use `.pck.hcl` files instead of  `.json`. The update readme is currently at a PR this is the link to the documentation to build this image: https://github.com/gem5/gem5-resources/blob/8b4955fa28b2c28bee4fb46e182137ca708d5d30/src/ubuntu-generic-diskimages/BUILDING.md. Sorry for the confusion, if you have any questions, I am happy to answer.

We use the `after_boot.sh` when we first login to a shell, so that we can run the readfile that is passed in gem5.

ncde:
thank you very much I will follow the building document

PR #37, opened 2024-05-29
resources: Add kernel extract to packer

Harshil2107:
- Now packer also extracts the kernel that is used in the diskimage to the output directory.

### DIFF
--- a/src/x86-ubuntu/BUILDING.md
+++ b/src/x86-ubuntu/BUILDING.md
@@ -23,6 +23,14 @@ You will see `qemu.initialize: Waiting for SSH to become available...` while the
 You can watch the installation with a VNC viewer.
 See [Troubleshooting](#troubleshooting) for more information.
 
+## Kernel
+
+A kernel is also extracted from the diskimage during the post-installation process.
+The latest headers and modules are installed using apt, before extracting the kernel using the `extract-vmlinux` script provided in ubuntu. The extracted kernel is placed at `/home/gem5/vmlinux-x86-ubuntu`.
+The extracted kernel does not have a version its name, but the kernel version is printed as before the extraction in `post-installation.sh` script. This extracted kernel can be used as a resource for gem5 simulations and is not limited to just be used with this diskiamge.
+
+The kernel is extracted using packer's file provisioner with `disrection=download` which would copy a file from the image to the host machine. The path specifying in the provisioner copies the file `/home/gem5/vmlinux-x86-ubuntu` to the output directory `disk-image`.
+
 ## Changes from the base Ubuntu 22.04 image
 
 - The default user is `gem5` with password `12345`.
--- a/src/x86-ubuntu/BUILDING.md
+++ b/src/x86-ubuntu/BUILDING.md
@@ -25,11 +25,11 @@ See [Troubleshooting](#troubleshooting) for more information.
 
 ## Kernel
 
-A kernel is also extracted from the diskimage during the post-installation process.
+A kernel is also extracted from the disk image during the post-installation process.
 The latest headers and modules are installed using apt, before extracting the kernel using the `extract-vmlinux` script provided in ubuntu. The extracted kernel is placed at `/home/gem5/vmlinux-x86-ubuntu`.
-The extracted kernel does not have a version its name, but the kernel version is printed as before the extraction in `post-installation.sh` script. This extracted kernel can be used as a resource for gem5 simulations and is not limited to just be used with this diskiamge.
+The extracted kernel does not have a version its name, but the kernel version is printed as before the extraction in `post-installation.sh` script. This extracted kernel can be used as a resource for gem5 simulations and is not limited to just be used with this disk image.
 
-The kernel is extracted using packer's file provisioner with `disrection=download` which would copy a file from the image to the host machine. The path specifying in the provisioner copies the file `/home/gem5/vmlinux-x86-ubuntu` to the output directory `disk-image`.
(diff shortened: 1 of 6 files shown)


src/x86-ubuntu/x86-ubuntu-24-04.pkr.hcl
```diff
@@ -78,4 +78,9 @@ build {
     expect_disconnect = true
   }
 
+  provisioner "file" {
+  source      = "/home/gem5/vmlinux-x86-6.8.0-31-generic"
+  destination = "./disk-image-ubuntu-24-04/vmlinux-x86-6.8.0-31-generic"
```
powerjg: This is going to break when Ubuntu updates the kernel (e.g., to 6.8.0-32). Maybe you should move `vmlinux-x86-$(uname -r)` to `vmlinux-x86-ubuntu` or something like that. There's probably a better way to do it..

powerjg (changes requested):

src/x86-ubuntu/x86-ubuntu-24-04.pkr.hcl
Harshil2107: SHould I change it in both places, or do we want the output kernel to specify the version?

powerjg: Yes, it needs to be done both places. However, we need a way to differentiate the kernels once we copy them out. I don't know the best way to do this.

Harshil2107: One thing could be that we print the kernel name in the post install, and that would show as an output. I will look into if we can have variables or programitcally name the files.

powerjg (approved):

powerjg (changes requested):
Oops. I changed my mind. Please update the `BUILDING.md` file to explain that the kernel is extracted and that you can use that kernel as a resource.

Think about how you would need to explain this to the next person who will be maintaining gem5 resources so they can update the disk and kernel. Or, more likely, think about what information you will need in two years to do this when 26.04 is released.

src/x86-ubuntu/BUILDING.md
```diff
@@ -23,6 +23,14 @@ You will see `qemu.initialize: Waiting for SSH to become available...` while the
 You can watch the installation with a VNC viewer.
 See [Troubleshooting](#troubleshooting) for more information.
 
+## Kernel
+
+A kernel is also extracted from the diskimage during the post-installation process.
+The latest headers and modules are installed using apt, before extracting the kernel using the `extract-vmlinux` script provided in ubuntu. The extracted kernel is placed at `/home/gem5/vmlinux-x86-ubuntu`.
+The extracted kernel does not have a version its name, but the kernel version is printed as before the extraction in `post-installation.sh` script. This extracted kernel can be used as a resource for gem5 simulations and is not limited to just be used with this diskiamge.
+
+The kernel is extracted using packer's file provisioner with `disrection=download` which would copy a file from the image to the host machine. The path specifying in the provisioner copies the file `/home/gem5/vmlinux-x86-ubuntu` to the output directory `disk-image`.
```
powerjg:
```suggestion
The kernel is extracted using packer's file provisioner with `direction=download` which would copy a file from the image to the host machine. The path specifying in the provisioner copies the file `/home/gem5/vmlinux-x86-ubuntu` to the output directory `disk-image`.
```

src/x86-ubuntu/BUILDING.md
```diff
@@ -23,6 +23,14 @@ You will see `qemu.initialize: Waiting for SSH to become available...` while the
 You can watch the installation with a VNC viewer.
 See [Troubleshooting](#troubleshooting) for more information.
 
+## Kernel
+
+A kernel is also extracted from the diskimage during the post-installation process.
```
powerjg:
```suggestion
A kernel is also extracted from the disk image during the post-installation process.
```

src/x86-ubuntu/scripts/post-installation.sh
```diff
@@ -75,5 +82,8 @@ elif [ -f /etc/netplan/00-installer-config.yaml ]; then
     mv /etc/netplan/00-installer-config.yaml /etc/netplan/00-installer-config.yaml.bak
     netplan apply
 fi
+# Disable systemd service that waits for network to be online
+systemctl disable systemd-networkd-wait-online.service
+systemctl mask systemd-networkd-wait-online.service
```
powerjg, line 87: The documentation is now wrong about how to re-enable networking. That needs to be updated.

src/x86-ubuntu/x86-ubuntu-24-04.pkr.hcl
```diff
@@ -78,4 +78,9 @@ build {
     expect_disconnect = true
   }
 
+  provisioner "file" {
+  source      = "/home/gem5/vmlinux-x86-ubuntu"
+  destination = "./disk-image-ubuntu-24-04/vmlinux-x86-ubuntu"
```
powerjg, line 83: Is this the only file that will be in the `disk-image-ubuntu-24-04` directory? Will the disk image be there? If so, how?

powerjg (changes requested):

src/x86-ubuntu/x86-ubuntu-24-04.pkr.hcl
Harshil2107: The disk image will be there as well, the output directory is different for the 24.04 image (https://github.com/gem5/gem5-resources/blob/6089e98242538e214802eb75cb908d9f400b783b/src/x86-ubuntu/x86-ubuntu-24-04.pkr.hcl#L41).

powerjg (approved):

PR #22, opened 2024-01-23
resources: Update GPU resources to Vega

abmerop:
This PR is part of [gem5/#703](https://github.com/gem5/gem5/issues/703) to update documentation. These commits do the following:

- Replace GCN3_X86 with VEGA_X86 for build and run commands.
- Replace gfx801 and gfx803 with gfx902 and gfx900 respectively.
- Update pre-built binary website paths to a v24-0 folder. These will have to be rebuilt to removed gfx8 targets and add gfx902.
- Update the docker version in READMEs to v23-1. This docker contains the additional packages and env variables to build HACC.
- Remove the hsa-agent-pkt application. This application has been broken for years and only builds with ROCm 1.6, so assuming it is no longer used.

PR #39, opened 2024-06-06
resources: Makefiles to build GPUFS apps using docker

abmerop:
Add Makefile.default for everything except HACC and DNNMark. These can be used with:

docker run <flags> ghcr.io/gem5/gpu-fs:latest make -f Makefile.default

### DIFF
--- a/src/gpu/pannotia/bc/BC.cpp
+++ b/src/gpu/pannotia/bc/BC.cpp
@@ -66,12 +66,16 @@
 #include "../graph_parser/util.h"
 #include "kernel.h"
 
-#ifdef GEM5_FUSION
+#if defined(GEM5_FUSION) || defined(GEM5_FS)
 #include <stdint.h>
 #include <gem5/m5ops.h>
 #endif
 
-#ifdef GEM5_FUSION
+#ifdef GEM5_FS
+#include <util/m5/src/m5_mmap.h>
+#endif
+
+#if defined(GEM5_FUSION) || defined(GEM5_FS)
 #define MAX_ITERS 150
 #else
 #include <stdint.h>
@@ -170,6 +174,12 @@ int main(int argc, char **argv)
     m5_work_begin(0, 0);
 #endif
 
+#ifdef GEM5_FS
+    m5op_addr = 0xFFFF0000;
+    map_m5_mem();
+    m5_work_begin_addr(0, 0);
+#endif
+
     // Copy data to device-side buffers
     err = hipMemcpy(row_d, csr->row_array, (num_nodes + 1) * sizeof(int), hipMemcpyHostToDevice);
     if (err != hipSuccess) {
@@ -262,6 +272,11 @@ int main(int argc, char **argv)
 
 #ifdef GEM5_FUSION
     m5_work_end(0, 0);
+#endif
+
+#ifdef GEM5_FS
+    m5_work_end_addr(0, 0);
+    unmap_m5_mem();
 #endif
 
     //timer2 = gettime();
--- a/src/gpu/pennant/src/HydroGPU.cxx
+++ b/src/gpu/pennant/src/HydroGPU.cxx
@@ -724,6 +724,8 @@ static __device__ void hydroCalcDtVolume(
 
 }
 
+
+#ifndef ATOMIC_MIN_DEFINED
 static __device__ double atomicMin(double* address, double val)
 {
     unsigned long long int* address_as_ull =
@@ -738,6 +740,7 @@ static __device__ double atomicMin(double* address, double val)
     } while (assumed != old);
     return __longlong_as_double(old);
 }
+#endif
 
 
 static __device__ void hydroFindMinDt(
--- a/src/gpu/heterosync/Makefile.default
+++ b/src/gpu/heterosync/Makefile.default
@@ -1,0 +1,22 @@
+EXECUTABLE := allSyncPrims-1kernel.default
+HIP_PATH ?= /opt/rocm
+
+SRC_DIR := src
+SRC := $(wildcard $(SRC_DIR)/*.hip.cpp)
+
+BIN_DIR := bin.default
+
+all: release-gfx9
+
+release-gfx9: $(SRC) | $(BIN_DIR)
+	$(HIP_PATH)/bin/hipcc -DGFX9 -Wno-unused-result --offload-arch=gfx900,gfx90a,gfx942 $(SRC) -o $(BIN_DIR)/$(EXECUTABLE)
+
+debug: $(SRC) | $(BIN_DIR)
+	$(HIP_PATH)/bin/hipcc -DDEBUG -g -O0 -Wno-unused-result --offload-arch=gfx900,gfx90a,gfx942 $(SRC) -o $(BIN_DIR)/$(EXECUTABLE).debug
+
+$(BIN_DIR):
+	mkdir -p $@
+
+clean:
+	rm -f $(BIN_DIR)/$(EXECUTABLE)
+	rmdir --ignore-fail-on-non-empty $(BIN_DIR)
--- a/src/gpu/hip-samples/Makefile.default
+++ b/src/gpu/hip-samples/Makefile.default
@@ -1,0 +1,22 @@
+HIP_PATH?= /opt/rocm
+HIPCC=$(HIP_PATH)/bin/hipcc
+
+SRC_DIR := src
+BIN_DIR := bin.default
+
+SOURCES := $(wildcard $(SRC_DIR)/*.cpp)
+EXECUTABLES := $(basename $(notdir $(SOURCES)))
+
+.PHONY: all clean
+
+all: $(EXECUTABLES)
+
+$(EXECUTABLES): %: $(SRC_DIR)/%.cpp | $(BIN_DIR)
+	$(HIPCC) -Wno-unused-result --offload-arch=gfx900,gfx90a,gfx942 $< -o $(BIN_DIR)/$@.default
+
+$(BIN_DIR):
+	mkdir -p $@
+
+clean:
+	rm -f $(BIN_DIR)/*.default
+	rmdir --ignore-fail-on-non-empty $(BIN_DIR)
--- a/src/gpu/lulesh/Makefile.default
+++ b/src/gpu/lulesh/Makefile.default
@@ -1,0 +1,12 @@
+BIN_DIR ?= ./bin.default
+HIP_PATH ?= /opt/rocm
+
+all: $(BIN_DIR)
+	$(HIP_PATH)/bin/hipcc src/lulesh.hip.cc -o $(BIN_DIR)/lulesh.default --offload-arch=gfx900,gfx90a,gfx942
+
+$(BIN_DIR):
+	mkdir -p $(BIN_DIR)
+
+clean:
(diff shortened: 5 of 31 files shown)


abmerop:
- This is on develop because it depends on https://github.com/gem5/gem5-resources/pull/22.
- We should discuss how to handle the README.md for this without it getting lost in SE mode details.
- Main changes:
   - hipcc path has changed since ROCm 4.
   - New hipcc has an absurd amounts of warning due to ignoring return value of hip*. Should we add -Wno-unused-result to the Makefile?
   - Make clean attempts to not clobber and SE mode makefile binaries but maybe there is a better approach (separate bin dir?).
- This does not include HACC or DNNMark:
   - I have *NO* idea how you are supposed to build HACC.
   - DNNMark requires changes to the GPUFS build docker image and some code changes but I eventually got it to work. I do not have a GPU machine with gtest to run the apps though.

mattsinc:
> his does not include HACC or DNNMark:
> 
>     * I have _NO_ idea how you are supposed to build HACC.
> 
>     * DNNMark requires changes to the GPUFS build docker image and some code changes but I eventually got it to work. I do not have a GPU machine with gtest to run the apps though.

In regards to HACC: you have to build its special Dockerfile first, then run make (although I vaguely recall you changed that in the past?).

Yes, we should add that warning flag.

Sure we could have separate bin directories for FS and SE mode.

src/gpu/pannotia/bc/Makefile.default
```diff
@@ -1,24 +1,22 @@
...
-OPTS = -O1
-
-HIP_PATH ?= /opt/rocm/hip
+HIP_PATH ?= /opt/rocm
 HIPCC = $(HIP_PATH)/bin/hipcc
 
+EXECUTABLE = bc_hip.default
+CPPSRC = BC.cpp 
+
+OPTS = -O3
```
mattsinc: are you seeing that BC passes with O3 on real GPUs now?  It was O1 before because of some sort of compilation error with ROCm 4 (maybe fixed?).  But this is a place where SE and FS may have to diverge -- since SE supports ROCm 4 at the moment.

src/gpu/pannotia/bc/Makefile.default
```diff
@@ -1,24 +1,22 @@
-CPPSRC = ../graph_parser/util.cpp
-CPPSRC = BC.cpp
-EXECUTABLE = bc_hip
-# BC has a compilation error in host code with ROCm 4.0, so O2 and O3 do not
-# pass even on real GPUs
-OPTS = -O1
-
-HIP_PATH ?= /opt/rocm/hip
+HIP_PATH ?= /opt/rocm
```
mattsinc, line 1: with Pannotia the Makefile.default was intended for compiling on the real GPU and Makefile.gem5-fusion was for gem5.  This change seems to be altering that, which means gem5's Makefiles would no longer be directly compatible with the non-gem5 Makefiles for Pannotia.  I guess we should discuss if that is a dealbreaker or not...

src/gpu/pannotia/color/Makefile.default
```diff
@@ -1,27 +1,30 @@
...
-	EXECUTABLE = $(BASEEXE)_max
-	CPPFILES += coloring_max.cpp
-else ifeq ($(VARIANT),MAXMIN)
-	EXECUTABLE = $(BASEEXE)_maxmin
-	CPPFILES += coloring_maxmin.cpp
-endif
+MAX_EXE = color_max.default
+MAXMIN_EXE = color_maxmin.default
 
+MAX_CPP = coloring_max.cpp
```
mattsinc, line 7: This isn't the end of the world, and it might even be for the better, but there was an explicit intent in forcing the user to pick what binary they wanted previously whereas now they get all of them.  I think I'm in favor of the change though ...

src/gpu/pennant/src/HydroGPU.cxx
```diff
@@ -724,6 +724,8 @@ static __device__ void hydroCalcDtVolume(
 
 }
 
+
+#ifndef ATOMIC_MIN_DEFINED
```
mattsinc, line 728: Shouldn't this be ifdef?

abmerop:
> In regards to HACC: you have to build its special Dockerfile first, then run make (although I vaguely recall you changed that in the past?).

I am trying to build on real hardware. Even after setting all the environment variables in the dockerfile it doesn't work.  I did not make that change btw, so not taking credit for it.

> Yes, we should add that warning flag.

Done

> Sure we could have separate bin directories for FS and SE mode.

Done

src/gpu/pannotia/color/Makefile.default
abmerop: What is the benefit of forcing them to pick?  If needed the default make target can change.  I strongly dislike the previous approach of needing to edit the makefile to change the variant though.

src/gpu/pannotia/bc/Makefile.default
abmerop: I tested all of these to make sure they build and run on hardware.  The path changed in ROCm 5 I think, so unless you have some ancient version of ROCm this will compile on a real GPU.  The gem5 makefiles specific to SE mode are untouched here and still use the old path.

src/gpu/pennant/src/HydroGPU.cxx
abmerop: I negated this to avoid touching SE mode make files.  atomicMin was added in some HIP version since ROCm 4, so keeping this results in duplicate symbols when linking.  I define this in makefile.default to build on modern ROCm versions.

src/gpu/pannotia/bc/Makefile.default
abmerop: Actually this one crashes in the parseCOO function in `_IO_fgets` which will be fun to debug.  So I wasn't actually able to test this.  Every other app works though and I would be surprised if whatever the bug was here still exists in ROCm 6.1.  I have built the rest of the benchmarks on real silicon and ran them and got correct results.

src/gpu/pannotia/color/Makefile.default
mattsinc: you didn't need to edit the Makefile, you just needed to set a variable on the command line (export VARIANT=...) but I agree picking is not necessary.

src/gpu/pannotia/bc/Makefile.default
mattsinc: Interesting, maybe I am misunderstanding.  I thought you intended the Makefile.default's to be the default Makefile for all gem5 GPUFS files?  If so, I had assumed you would need to be including m5ops and the like somewhere in the linkage for the Makefile.default's, but since that isn't the case maybe you intend Makefile.default to be for running on the real GPU?  Sorry if I'm missing something dumb here.

src/gpu/pannotia/bc/Makefile.default
mattsinc: Should we revert to O1 then?  I believe this is the same bug I was getting ...

src/gpu/pannotia/bc/Makefile.default
abmerop: Oh, I see. So this does work with -O1.  The previous comment was a bit misleading because it is not a compile error and seems unrelated to GPUs since it crashes in fgets().  I changed this back to -O1 and reworded the comment.

abmerop: The goal of GPUFS is to run unmodified binaries, so the makefile for GPUFS and makefile for a real system would be identical

mattsinc: But how are you handling the m5ops in those files?  Those will not work on the real GPU.

abmerop:
I've added `Makefile.gpufs` to handle the m5ops.  These only exist in pannotia so only pannotia has these.  `Makefile.default` now builds for the host only.  For all of the other resources, `Makefile.default` should be used to run in GPUFS.

One other thing, I am not sure about what to do with the buildall.sh script in pannotia. It seems to only build the gem5-fusion target and there are now three targets (gem5-fusion, default, gpufs).

mattsinc:
> I've added `Makefile.gpufs` to handle the m5ops. These only exist in pannotia so only pannotia has these. `Makefile.default` now builds for the host only. For all of the other resources, `Makefile.default` should be used to run in GPUFS.
> 
> One other thing, I am not sure about what to do with the buildall.sh script in pannotia. It seems to only build the gem5-fusion target and there are now three targets (gem5-fusion, default, gpufs).

we could rename buildall.sh to buildall-gem5fusion.sh?  And add buildall-gpufs.sh?

src/gpu/pannotia/bc/Makefile.default
mattsinc: Just making sure you didn't miss my reply here

abmerop:
> we could rename buildall.sh to buildall-gem5fusion.sh? And add buildall-gpufs.sh?

The current behavior is `./buildall.sh` will build Makefile.default, and `./buildall.sh foo` will build gem5-fusion with `extraparams=foo`. We could probably modify it to allow you to select the target. As it is right now, it is selecting gpufs build for some reason. I don't know how make selects which target to build.

What i'd propose is making the `$1` be the makefile target (default, gem5-fusion, or gpufs) and `$2` be the extraparam.  If no option it would be either an error or Makefile.default.

mattsinc:
> > we could rename buildall.sh to buildall-gem5fusion.sh? And add buildall-gpufs.sh?
> 
> The current behavior is `./buildall.sh` will build Makefile.default, and `./buildall.sh foo` will build gem5-fusion with `extraparams=foo`. We could probably modify it to allow you to select the target. As it is right now, it is selecting gpufs build for some reason. I don't know how make selects which target to build.
> 
> What i'd propose is making the `$1` be the makefile target (default, gem5-fusion, or gpufs) and `$2` be the extraparam. If no option it would be either an error or Makefile.default.

This makes sense to me

src/gpu/pannotia/bc/BC.cpp
```diff
@@ -170,6 +174,11 @@ int main(int argc, char **argv)
     m5_work_begin(0, 0);
 #endif
 
+#ifdef GEM5_FS
+    map_m5_mem();
```
mattsinc, line 179: Do you need to add the "magic" address here too?

(same question for all Pannotia apps, but not going to add every single one as a comment)

mattsinc (changes requested):

src/gpu/pannotia/bc/Makefile.default
mattsinc: I see you updated this, so resolving

src/gpu/pannotia/bc/BC.cpp
abmerop: Sigh.  Yes.  It looks like the default was changed to zero so I guess I have to build in the m5_mmap file which I'm going to guess will cause a duplicate linker error because it's linked into the libm5.a...  There doesn't appear to be a way to set the value when scons'ing the library either.  I don't know how this was ever supposed to work.

abmerop: Figured that out and made the buildall change as well, which involved a few other fixes.

mattsinc (approved):

PR #41, opened 2024-06-25
resources: add build option to build riscv 24.04 image.

Harshil2107:
- This change also removes http directory as we are not using it.


powerjg (approved):

powerjg:
Please put the URLs and the names of the resources (and workloads, if appropriate) in a comment in this PR

PR #43, opened 2024-07-11
resources: Change pannotia fstream from fopen

mattsinc:
A recent commit (b17de6924e) changed Pannotia's parser to use fstream instead of fgets. However, this change introduces a "bad_alloc" error with some versions of ROCm.  To resolve this, change Pannotia's parse to use fopen instead, which works for both GPUSE and GPUFS in our testing.


mattsinc:
This work was done by @jebraun3 but I'm pushing as James cannot at the moment and the Pannotia tests cannot run in weekly until this is resolved (I think).

mattsinc:
Something got messed up when I attempted to update commit message, closing and reopening.

PR #44, opened 2024-07-11
resources: Change pannotia to fopen from fstream

mattsinc:
A recent commit (b17de69) changed Pannotia's parser to use fstream instead of fgets. However, this change introduces a "bad_alloc" error with some versions of ROCm. To resolve this, change Pannotia's parse to use fopen instead, which works for both GPUSE and GPUFS in our testing.

The was done by @jebraun3 but I am pushing to resolve problem affecting all Pannotia (weekly) tests.

### DIFF
--- a/src/gpu/pannotia/graph_parser/parse.cpp
+++ b/src/gpu/pannotia/graph_parser/parse.cpp
@@ -2,6 +2,7 @@
  *                                                                                  *
  * Copyright � 2014 Advanced Micro Devices, Inc.                                    *
  * Copyright (c) 2015 Mark D. Hill and David A. Wood                                *
+ * Copyright (c) 2024 James Braun and Matthew D. Sinclair                           *
  * All rights reserved.                                                             *
  *                                                                                  *
  * Redistribution and use in source and binary forms, with or without               *
@@ -119,24 +120,25 @@ ell_array *csr2ell(csr_array *csr, int num_nodes, int num_edges, int fill)
 
 csr_array *parseMetis(char* tmpchar, int *p_num_nodes, int *p_num_edges, bool directed)
 {
-
     int cnt = 0;
     unsigned int lineno = 0;
-    char *line = (char *)malloc(8192);
+    char *line = NULL;
     int num_edges = 0, num_nodes = 0;
+    size_t len = 128;
+    ssize_t nread;
+    FILE *stream;
 
-    std::fstream fp(tmpchar, std::ios_base::in);
-    CooTuple *tuple_array = NULL;
-
-    if (!fp.good()) {
+    stream = fopen(tmpchar, "r");
+    if (stream == NULL) {
         fprintf(stderr, "Error when opening file: %s\n", tmpchar);
         exit(1);
     }
 
+    CooTuple *tuple_array = NULL;
+
     printf("Opening file: %s\n", tmpchar);
 
-    while (!fp.eof()) {
-        fp.getline(line, 8192);
+    while ((nread = getline(&line, &len, stream)) != -1) {
         int head, tail, weight = 0;
         CooTuple temp;
 
@@ -221,12 +223,11 @@ csr_array *parseMetis(char* tmpchar, int *p_num_nodes, int *p_num_edges, bool di
     csr->col_array = col_array;
     csr->data_array = data_array;
 
-    fp.close();
     free(tuple_array);
     free(line);
+    fclose(stream);
 
     return csr;
-
 }
 
 
@@ -234,22 +235,23 @@ csr_array *parseCOO(char* tmpchar, int *p_num_nodes, int *p_num_edges, bool dire
 {
     int cnt = 0;
     unsigned int lineno = 0;
-    char line[128], sp[2], a, p;
+    char sp[2], a, p;
+    char *line = NULL;
     int num_nodes = 0, num_edges = 0;
+    size_t len = 128;
+    ssize_t nread;
+    FILE *stream;
 
-    std::fstream fp(tmpchar, std::ios_base::in);
-    CooTuple *tuple_array = NULL;
-
-    if (!fp.good()) {
+    stream = fopen(tmpchar, "r");
+    if (stream == NULL) {
         fprintf(stderr, "Error when opening file: %s\n", tmpchar);
         exit(1);
     }
+    CooTuple *tuple_array = NULL;
 
     printf("Opening file: %s\n", tmpchar);
 
-    while (!fp.eof()) {
-        fp.getline(line, 100);
-
+    while ((nread = getline(&line, &len, stream)) != -1) {
         int head, tail, weight;
         switch (line[0]) {
         case 'c':
@@ -270,7 +272,6 @@ csr_array *parseCOO(char* tmpchar, int *p_num_nodes, int *p_num_edges, bool dire
             printf("Read from file: num_nodes = %d, num_edges = %d\n", num_nodes, num_edges);
             tuple_array = (CooTuple *)malloc(sizeof(CooTuple) * num_edges);
             break;
-
         case 'a':
             sscanf(line, "%c %d %d %d", &a, &head, &tail, &weight);
             if (tail == head) printf("reporting self loop\n");
@@ -325,7 +326,8 @@ csr_array *parseCOO(char* tmpchar, int *p_num_nodes, int *p_num_edges, bool dire
 
     row_array[row_cnt] = idx;
 
-    fp.close();
+    free(line);
+    fclose(stream);
     free(tuple_array);
 
     csr_array *csr = (csr_array *)malloc(sizeof(csr_array));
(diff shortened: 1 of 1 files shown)


mattsinc:
@abmerop : I checked and FW, which uses a different parsing function, appears to already use fopen: https://github.com/gem5/gem5-resources/blob/stable/src/gpu/pannotia/fw/parse.cpp#L35

So I think it doesn't need to be changed ...

abmerop:
It seems that #24 fixes GPUFS but breaks SE (the current issue). This effectively reverts #24 and breaks GPUFS but fixes SE.  I believe FW works on both.  We should try and figure out what FW is doing and replicate that? 

For reference getline() hangs / crashes in ROCm 6+.  The fgets() in FW seems to work fine on both though.

abmerop (changes requested):
gfets instead of getline .

mattsinc:
> It seems that #24 fixes GPUFS but breaks SE (the current issue). This effectively reverts #24 and breaks GPUFS but fixes SE. I believe FW works on both. We should try and figure out what FW is doing and replicate that?
> 
> For reference getline() hangs / crashes in ROCm 6+. The fgets() in FW seems to work fine on both though.

I don't think changing back to fgets (which I think is what you're asking) makes sense because that is what you originally changed it from, right?

Maybe this is a fundamental ROCm 4 vs. 6 thing and we need to have an ifdef controlling which is used for now?

abmerop:
> > It seems that #24 fixes GPUFS but breaks SE (the current issue). This effectively reverts #24 and breaks GPUFS but fixes SE. I believe FW works on both. We should try and figure out what FW is doing and replicate that?
> > For reference getline() hangs / crashes in ROCm 6+. The fgets() in FW seems to work fine on both though.
> 
> I don't think changing back to fgets (which I think is what you're asking) makes sense because that is what you originally changed it from, right?
> 
> Maybe this is a fundamental ROCm 4 vs. 6 thing and we need to have an ifdef controlling which is used for now?

I believe one calls the other.  Let me try this again because I just noticed you are getline'ing into a nullptr. Maybe it works if `line` is allocated.

mattsinc:
> > > It seems that #24 fixes GPUFS but breaks SE (the current issue). This effectively reverts #24 and breaks GPUFS but fixes SE. I believe FW works on both. We should try and figure out what FW is doing and replicate that?
> > > For reference getline() hangs / crashes in ROCm 6+. The fgets() in FW seems to work fine on both though.
> > 
> > 
> > I don't think changing back to fgets (which I think is what you're asking) makes sense because that is what you originally changed it from, right?
> > Maybe this is a fundamental ROCm 4 vs. 6 thing and we need to have an ifdef controlling which is used for now?
> 
> I believe one calls the other. Let me try this again because I just noticed you are getline'ing into a nullptr. Maybe it works if `line` is allocated.

I believe we passed a NULL string into getline because that was what the man page recommended, but I can't seem to find it.  @jebraun3 if you see this and remember, your input would be appreciated :)

My recollection though was if a string exceeded the allocated char * size, then it was causing a failure and James and I had found something online saying that a NULL char * should be passed in instead ...

The C++ page on getline is not helpful: https://cplusplus.com/reference/string/string/getline/, although it does show passing a string in instead of char * -- maybe we should switch and see if the C++ string library handles the issue under the hood?

abmerop:
> > > 
> > > 
> > > I don't think changing back to fgets (which I think is what you're asking) makes sense because that is what you originally changed it from, right?
> > > Maybe this is a fundamental ROCm 4 vs. 6 thing and we need to have an ifdef controlling which is used for now?
> > 
> > 
> > I believe one calls the other. Let me try this again because I just noticed you are getline'ing into a nullptr. Maybe it works if `line` is allocated.
> 
> I believe we passed a NULL string into getline because that was what the man page recommended, but I can't seem to find it. @jebraun3 if you see this and remember, your input would be appreciated :)
> 
> My recollection though was if a string exceeded the allocated char * size, then it was causing a failure and James and I had found something online saying that a NULL char * should be passed in instead ...
> 
> The C++ page on getline is not helpful: https://cplusplus.com/reference/string/string/getline/, although it does show passing a string in instead of char * -- maybe we should switch and see if the C++ string library handles the issue under the hood?

I see. Allocating also doesn't work unfortunately.  So either we need to find a way to read a file that works for both modes or ifdef for GEM5_FUSION vs. GEM5_FS. It might make sense to have completely separate files and ifdef the include given the amount of changes.

mattsinc:
> > > > I don't think changing back to fgets (which I think is what you're asking) makes sense because that is what you originally changed it from, right?
> > > > Maybe this is a fundamental ROCm 4 vs. 6 thing and we need to have an ifdef controlling which is used for now?
> > > 
> > > 
> > > I believe one calls the other. Let me try this again because I just noticed you are getline'ing into a nullptr. Maybe it works if `line` is allocated.
> > 
> > 
> > I believe we passed a NULL string into getline because that was what the man page recommended, but I can't seem to find it. @jebraun3 if you see this and remember, your input would be appreciated :)
> > My recollection though was if a string exceeded the allocated char * size, then it was causing a failure and James and I had found something online saying that a NULL char * should be passed in instead ...
> > The C++ page on getline is not helpful: https://cplusplus.com/reference/string/string/getline/, although it does show passing a string in instead of char * -- maybe we should switch and see if the C++ string library handles the issue under the hood?
> 
> I see. Allocating also doesn't work unfortunately. So either we need to find a way to read a file that works for both modes or ifdef for GEM5_FUSION vs. GEM5_FS. It might make sense to have completely separate files and ifdef the include given the amount of changes.

If you change it to a string instead of a char *, does it work on your end?

abmerop:
This seems to be another bizarre compiler bug involving optimization. I made the following `Makefile.default` change to get `mis` to work natively.  The same GPUFS bug happens natively and it is faster to iterate natively, hence starting with that.  

**My suggestion is to go with this PR as it is now and add a commit changing all of the `Makefile.default` and `Makefile.gpufs` files.** I will test other pannotia bmks while waiting for feedback.

```--- a/src/gpu/pannotia/mis/Makefile.default
+++ b/src/gpu/pannotia/mis/Makefile.default
@@ -10,7 +10,9 @@ BIN_DIR ?= ./bin.default
 all: $(BIN_DIR)/$(EXECUTABLE)

 $(BIN_DIR)/$(EXECUTABLE): $(CPPSRC) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(BIN_DIR)
-       $(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(CPPSRC) -o $(BIN_DIR)/$(EXECUTABLE)
+       $(HIPCC) -O0 --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/parse.cpp -o $(BIN_DIR)/parse.o -c
+       $(HIPCC) -O0 --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/util.cpp -o $(BIN_DIR)/util.o -c
+       $(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) $(BIN_DIR)/parse.o $(BIN_DIR)/util.o $(CPPSRC) -o $(BIN_DIR)/$(EXECUTABLE)

 $(BIN_DIR):
        mkdir -p $(BIN_DIR)
```

Same change to `Makefile.gpufs`. **No** changes to `Makefile.gem5-fusion` (SE mode). Basically, build the parser with `-O0` and GPU code with `-O3`.  I am fine with this solution as CPU code is not ROI.

Interestingly enough, I was also able to get it to work with -O3 by enabling VERBOSE.  The extra printf added there seems to be enough to fix what appears to be a race condition between a futex being created and used. But that is getting outside of my wheelhouse.

mattsinc:
> This seems to be another bizarre compiler bug involving optimization. I made the following `Makefile.default` change to get `mis` to work natively. The same GPUFS bug happens natively and it is faster to iterate natively, hence starting with that.
> 
> **My suggestion is to go with this PR as it is now and add a commit changing all of the `Makefile.default` and `Makefile.gpufs` files.** I will test other pannotia bmks while waiting for feedback.
> 
> ```
> +++ b/src/gpu/pannotia/mis/Makefile.default
> @@ -10,7 +10,9 @@ BIN_DIR ?= ./bin.default
>  all: $(BIN_DIR)/$(EXECUTABLE)
> 
>  $(BIN_DIR)/$(EXECUTABLE): $(CPPSRC) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(BIN_DIR)
> -       $(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(CPPSRC) -o $(BIN_DIR)/$(EXECUTABLE)
> +       $(HIPCC) -O0 --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/parse.cpp -o $(BIN_DIR)/parse.o -c
> +       $(HIPCC) -O0 --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/util.cpp -o $(BIN_DIR)/util.o -c
> +       $(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) $(BIN_DIR)/parse.o $(BIN_DIR)/util.o $(CPPSRC) -o $(BIN_DIR)/$(EXECUTABLE)
> 
>  $(BIN_DIR):
>         mkdir -p $(BIN_DIR)
> ```
> 
> Same change to `Makefile.gpufs`. **No** changes to `Makefile.gem5-fusion` (SE mode). Basically, build the parser with `-O0` and GPU code with `-O3`. I am fine with this solution as CPU code is not ROI.
> 
> Interestingly enough, I was also able to get it to work with -O3 by enabling VERBOSE. The extra printf added there seems to be enough to fix what appears to be a race condition between a futex being created and used. But that is getting outside of my wheelhouse.

Just to be clear, you want me to make this change to the Makefile's as part of this commit (or an additional commit in this pull request)?

abmerop:
I can make the change as an additional commit or a new PR.  I'll approve and if you merge i'll take that as I should make another PR.

abmerop (approved):

mattsinc:
> I can make the change as an additional commit or a new PR. I'll approve and if you merge i'll take that as I should make another PR.

ok, either way would have been fine, but will go with your plan

abmerop:
This is better actually:  will fix our CI and if gem5 tests them, those too.

abmerop:
One final thought: Our SE mode CI tests are passing now, thanks for the PR.

mattsinc:
> One final thought: Our SE mode CI tests are passing now, thanks for the PR.

Progress, great -- nice job @jebraun3 !

PR #15, opened 2023-11-12
resources: updated pannotia documentation for compile commands

KaiBatley:
The docker command for compiling pannotia tests needs to have visibility to m5ops.

PR #45, opened 2024-07-19
resources: Update pannotia Makefiles for GPUFS/HW

abmerop:
There is an issue with the compiler used with GPUFS and hardware that causes a hang or crash in fstream. PR #44 changed fstream to fopen as fstream has similar issues in ROCm 4.0. To compromise and support both versions, this PR updates the GPUFS and hardware (Makefile.default) Makefiles to build the input parser with optimizations off (-O0) which "solves" the issue. Since we typically do not care about simulation time of reading input files, this seems like a decent compromise.

The GPUFS Makefiles also change the GEM5_PATH environment variable to GEM5_ROOT to be consistent with each Makefile.gem5-fusion.

The BC application is unchanged as it implements its open parser in its header file which cannot be compiled to an object file without refactoring the benchmark.

### DIFF
--- a/src/gpu/pannotia/mis/Makefile.default
+++ b/src/gpu/pannotia/mis/Makefile.default
@@ -10,7 +10,10 @@ BIN_DIR ?= ./bin.default
 all: $(BIN_DIR)/$(EXECUTABLE)
 
 $(BIN_DIR)/$(EXECUTABLE): $(CPPSRC) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(BIN_DIR)
-	$(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(CPPSRC) -o $(BIN_DIR)/$(EXECUTABLE)
+	# There is a futex issue in hipcc with the graph parser. Compile without opts until this is fixed.
+	$(HIPCC) -O0 --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/parse.cpp -o $(BIN_DIR)/parse.o -c
+	$(HIPCC) -O0 --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/util.cpp -o $(BIN_DIR)/util.o -c
+	$(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) $(BIN_DIR)/parse.o $(BIN_DIR)/util.o $(CPPSRC) -o $(BIN_DIR)/$(EXECUTABLE)
 
 $(BIN_DIR):
 	mkdir -p $(BIN_DIR)
--- a/src/gpu/pannotia/pagerank/Makefile.default
+++ b/src/gpu/pannotia/pagerank/Makefile.default
@@ -7,17 +7,22 @@ SPMV_EXE = pagerank_spmv.default
 DEF_CPP = pagerank.cpp
 SPMV_CPP = pagerank_spmv.cpp
 
-CPPSRC = ../graph_parser/util.cpp ../graph_parser/parse.cpp
 OPTS = -O3 -Wno-unused-result
 BIN_DIR ?= ./bin.default
 
 all: $(BIN_DIR)/$(DEF_EXE) $(BIN_DIR)/$(SPMV_EXE)
 
 $(BIN_DIR)/$(DEF_EXE): $(DEF_CPP) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(BIN_DIR)
-	$(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) $(DEF_CPP) $(CPPSRC) -o $(BIN_DIR)/$(DEF_EXE)
+	# There is a futex issue in hipcc with the graph parser. Compile without opts until this is fixed.
+	$(HIPCC) -O0 --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/parse.cpp -o $(BIN_DIR)/parse.o -c
+	$(HIPCC) -O0 --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/util.cpp -o $(BIN_DIR)/util.o -c
+	$(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) $(BIN_DIR)/parse.o $(BIN_DIR)/util.o $(DEF_CPP) -o $(BIN_DIR)/$(DEF_EXE)
 
 $(BIN_DIR)/$(SPMV_EXE): $(SPMV_CPP) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(BIN_DIR)
-	$(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) $(SPMV_CPP) $(CPPSRC) -o $(BIN_DIR)/$(SPMV_EXE)
+	# There is a futex issue in hipcc with the graph parser. Compile without opts until this is fixed.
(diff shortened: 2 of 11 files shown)


src/gpu/pannotia/mis/Makefile.default
```diff
@@ -10,7 +10,10 @@ BIN_DIR ?= ./bin.default
 all: $(BIN_DIR)/$(EXECUTABLE)
 
 $(BIN_DIR)/$(EXECUTABLE): $(CPPSRC) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(BIN_DIR)
-	$(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(CPPSRC) -o $(BIN_DIR)/$(EXECUTABLE)
+    # There is a futex issue in hipcc with the graph parser. Compile without opts until this is fixed.
```
mattsinc: indented at different levels? (lines 13 vs. 14-16)

src/gpu/pannotia/pagerank/Makefile.default
```diff
@@ -7,17 +7,22 @@ SPMV_EXE = pagerank_spmv.default
...
 
-CPPSRC = ../graph_parser/util.cpp ../graph_parser/parse.cpp
 OPTS = -O3 -Wno-unused-result
 BIN_DIR ?= ./bin.default
 
 all: $(BIN_DIR)/$(DEF_EXE) $(BIN_DIR)/$(SPMV_EXE)
 
 $(BIN_DIR)/$(DEF_EXE): $(DEF_CPP) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(BIN_DIR)
-	$(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) $(DEF_CPP) $(CPPSRC) -o $(BIN_DIR)/$(DEF_EXE)
+    # There is a futex issue in hipcc with the graph parser. Compile without opts until this is fixed.
```
mattsinc: same here

src/gpu/pannotia/sssp/Makefile.default
```diff
@@ -7,17 +7,22 @@ ELL_EXE = sssp_ell.default
...
 
-CPPSRC = ../graph_parser/util.cpp ../graph_parser/parse.cpp
 OPTS = -O3 -Wno-unused-result
 BIN_DIR ?= ./bin.default
 
 all: $(BIN_DIR)/$(CSR_EXE) $(BIN_DIR)/$(ELL_EXE)
 
 $(BIN_DIR)/$(CSR_EXE): $(CSR_CPP) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(BIN_DIR)
-	$(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) $(CSR_CPP) $(CPPSRC) -o $(BIN_DIR)/$(CSR_EXE)
+    # There is a futex issue in hipcc with the graph parser. Compile without opts until this is fixed.
```
mattsinc: same here

src/gpu/pannotia/sssp/Makefile.default
```diff
@@ -7,17 +7,22 @@ ELL_EXE = sssp_ell.default
...
+	$(HIPCC) -O0 --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/parse.cpp -o $(BIN_DIR)/parse.o -c
+	$(HIPCC) -O0 --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) ../graph_parser/util.cpp -o $(BIN_DIR)/util.o -c
+	$(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) $(BIN_DIR)/parse.o $(BIN_DIR)/util.o $(CSR_CPP) -o $(BIN_DIR)/$(CSR_EXE)
 
 $(BIN_DIR)/$(ELL_EXE): $(ELL_CPP) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(BIN_DIR)
-	$(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) $(ELL_CPP) $(CPPSRC) -o $(BIN_DIR)/$(ELL_EXE)
+    # There is a futex issue in hipcc with the graph parser. Compile without opts until this is fixed.
```
mattsinc: same here

src/gpu/pannotia/sssp/Makefile.gpufs
```diff
@@ -7,18 +7,23 @@ ELL_EXE = sssp_ell.gpufs
...
+OPTS = -O3 -Wno-unused-result -DGEM5_FS -I$(GEM5_ROOT)/include -I$(GEM5_ROOT)
 BIN_DIR ?= ./bin.gpufs
-LDFLAGS += -L$(GEM5_PATH)/util/m5/build/x86/out -lm5
+LDFLAGS += -L$(GEM5_ROOT)/util/m5/build/x86/out -lm5
 
 all: $(BIN_DIR)/$(CSR_EXE) $(BIN_DIR)/$(ELL_EXE)
 
 $(BIN_DIR)/$(CSR_EXE): $(CSR_CPP) ../graph_parser/parse.cpp ../graph_parser/util.cpp $(BIN_DIR)
-	$(HIPCC) $(OPTS) --offload-arch=gfx900,gfx90a,gfx942 $(CXXFLAGS) $(CSR_CPP) $(CPPSRC) -o $(BIN_DIR)/$(CSR_EXE) $(LDFLAGS)
+    # There is a futex issue in hipcc with the graph parser. Compile without opts until this is fixed.
```
mattsinc: same here (going to stop commenting on this here)

mattsinc (changes requested):
Two questions:

1.  Does BC just not work in this version?
2.  Is there a wrapper Makefile above these that sets GEM5_ROOT, or do we expect users to set it?

abmerop:
> Two questions:
> 
> 1. Does BC just not work in this version?
> 2. Is there a wrapper Makefile above these that sets GEM5_ROOT, or do we expect users to set it?

BC works.  The current solution is to build with `-O1`.

GEM5_ROOT needs to be set by the user.  I guess a reasonable default could be `../../../../gem5` though (assumes a base directory containing gem5 repo and gem5-resources repo).

mattsinc:
> > Two questions:
> > 
> > 1. Does BC just not work in this version?
> > 2. Is there a wrapper Makefile above these that sets GEM5_ROOT, or do we expect users to set it?
> 
> BC works. The current solution is to build with `-O1`.
> 
> GEM5_ROOT needs to be set by the user. I guess a reasonable default could be `../../../../gem5` though (assumes a base directory containing gem5 repo and gem5-resources repo).

Right, the old Makefiles used:

GEM5_PATH ?= ...

To allow the users to have this set already, but guess something otherwise.  I think your change makes it so this is essentially not done now, and the user needs to look at this and guess?

abmerop:
> 
> Right, the old Makefiles used:
> 
> GEM5_PATH ?= ...
> 
> To allow the users to have this set already, but guess something otherwise. I think your change makes it so this is essentially not done now, and the user needs to look at this and guess?

I added a default `GEM5_ROOT ?= ../../../../../gem5` .  I assume you meant `GEM5_ROOT` rather than `GEM5_PATH`. I am using the same variable as in each `Makefile.gem5-fusion`.

mattsinc:
> > Right, the old Makefiles used:
> > GEM5_PATH ?= ...
> > To allow the users to have this set already, but guess something otherwise. I think your change makes it so this is essentially not done now, and the user needs to look at this and guess?
> 
> I added a default `GEM5_ROOT ?= ../../../../../gem5` . I assume you meant `GEM5_ROOT` rather than `GEM5_PATH`. I am using the same variable as in each `Makefile.gem5-fusion`.

The old versions were using GEM5_PATH instead, but GEM5_ROOT is ok.

mattsinc (approved):

abmerop:
> The old versions were using GEM5_PATH instead, but GEM5_ROOT is ok.

It looks like the READMEs were updated in #15 to say GEM5_ROOT, so I will keep that

mattsinc:
> > The old versions were using GEM5_PATH instead, but GEM5_ROOT is ok.
> 
> It looks like the READMEs were updated in #15 to say GEM5_ROOT, so I will keep that

That was probably not intentional, but is fine.

PR #48, opened 2024-08-19
DRAFT: resources: Add gem5_bridge driver to x86-ubuntu build & boot

nkrim:
Adds commands to `scripts/post-installation.sh` to make the driver during disk building. The Makefile can take arguments to modify the base address for the gem5ops MMIO range by putting `GEM5OPS_BASE=<base_phys_addr>` at the end of the make command.

Adds commands to `files/gem5_init.sh` to detect and add the driver after kernel startup but just before the first `gem5-bridge exit`.

Relates to https://github.com/gem5/gem5/pull/1480


Harshil2107:
I am closing this PR as gem5 bridge driver is already added to the current disk images via other PRs.

PR #52, opened 2024-09-04
resources: ARM Ubuntu 24.04 disk image with SPEC 2017

erin-le:
This pr adds the files needed to create an ARM Ubuntu 24.04 disk image with an installation of SPEC 2017.

I still need to update the README to reflect the current method of creating disk images.


src/arm-ubuntu-spec-2017/README.md
```diff
@@ -0,0 +1,201 @@
+---
+title: Linux arm-ubuntu image
+tags:
+    - arm
+    - fullsystem
+layout: default
+permalink: resources/arm-ubuntu
+shortdoc: >
+    Resources to build a generic arm-ubuntu disk image.
+author: ["Hoa Nguyen", "Kaustav Goswami"]
```
powerjg, line 10: You should include yourself here

src/arm-ubuntu-spec-2017/arm-ubuntu-24-04.pkr.hcl
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  provisioner "file" {
+    destination = "/home/gem5/"
+    source      = "files/after_boot.sh"
+  }
+
+  provisioner "file" {
+    destination = "/home/gem5/"
+    source      = "/home/bees/gem5-bootcamp-private/cpu2017-1.1.0.iso"
```
powerjg, line 87: You can probably have this be part of the `initialize` block by loading it as a cdrom drive.

src/arm-ubuntu-spec-2017/build.sh
```diff
@@ -0,0 +1,27 @@
...
+    exit 1
+fi
+
+# Store the configuration file name from the command line argument
+config_file="$1"
+
+# Check if the specified configuration file exists
+if [ -f "$config_file" ]; then
+    # Install the needed plugins
+    PACKER_LOG=1 ./packer init "$config_file"
```
powerjg, line 21: Make sure to remove this logging in the final version

src/arm-ubuntu-spec-2017/files/after_boot.sh
powerjg, line 1: We probably want to simplify this file for the spec suite. We want to simply say what spec workload to run. Maybe we should only accept parameters to the `runspec` script.

src/arm-ubuntu-spec-2017/files/exit.sh
powerjg, line 1: We likely don't need this file anymore.

powerjg (commented):
Is the thought that we will always build this disk image from scratch? If so, why? The alternative is to do what we do in npb and gapbs now which is to use the base ubuntu image and build on top of it. I don't have a strong feeling either way, but I would like to understand your thought process.

powerjg:
Another to do item before merging this is to figure out how to add ROIs. Zhantong (I can't find her github handle) should be able to help with that.

erin-le:
> Is the thought that we will always build this disk image from scratch? If so, why? The alternative is to do what we do in npb and gapbs now which is to use the base ubuntu image and build on top of it. I don't have a strong feeling either way, but I would like to understand your thought process.

To be honest, there wasn't any particular thought process; building it from scratch was just the only way I could figure out how to do it. If the NPB and GAPBS images are built on top of the base disk image, it'll be more consistent and probably be better to build the SPEC disk images the same way.

Harshil2107:
Just a thing to note, the size of the SPEC image is around 11.5 GB but the base images we have are only 5GB. We have two options:

- make the spec image from scratch
- Use the base image but define the size to be bigger in the packer script (this will resize the disk image but not the file system/ partition on the image), then in boot command we add the commands to resize the partition to match the updated size of the disk. This seems to me a more hacky way to do it (It should work in theory but I have not tested this re size method).

Harshil2107:
you can add the following to the post install script to resize the partition and filesystem.  You would have to resize the disk by updating the `disk_size` field in the packer file to be bigger that the size of the current disk.

```
echo "Starting partition resizing..."

# Run fdisk to resize /dev/vda2
(
  echo d        # Delete partition
  echo 2        # Select partition 2 (assuming it's /dev/vda2)
  echo n        # Create new partition
  echo p        # Primary partition
  echo 2        # Partition number 2
  echo          # Default first sector (same as before)
  echo          # Default last sector (expand to full size)
  echo N        # Keep the existing ext4 signature
  echo w        # Write changes
) | sudo fdisk /dev/vda

# Resize the filesystem to use the expanded partition
sudo resize2fs /dev/vda2

echo "Partition resizing completed successfully!"
```

powerjg:
Two comments:
1. Did you look at using `parted`? That has a "resize" command.
2. What partition are you deleting?

In the comments, you should write what `print` looks like each time.

Harshil2107:
> Two comments:
> 
> 1. Did you look at using `parted`? That has a "resize" command.

I looked into it a little, when I try to resize, it gives a warning that the partition is already in use, as that was stalling the script, I will try to echo the replies in order and see if it works. Also parted is not on the disk by default, so we will have to install it before we update the partition.

> 2. What partition are you deleting?

I am deleting partition 2 (/dev/vda2 as we have two partitions for the arm image one for boot and one root), and creating a new partition 2 with full size.

> In the comments, you should write what `print` looks like each time.

I dont understand this, do you mean the output from fdisk for each input I am echoing in?

erin-le:
I am not currently working on this, so it is being closed. I will reopen it if it becomes relevant again.

PR #53, opened 2024-09-04
resources: add Arm SPEC 2006 Ubuntu 24.04 disk image

erin-le:
This adds files used to make an Arm Ubuntu 24.04 disk image with a working installation of SPEC 2006. Please note that the installation process modifies a number of files in the SPEC 2006 image based on the necessary changes listed here: https://github.com/studyztp/some-notes/blob/main/SPEC2006_RISCV_ubuntu20.04.4_issue_note.md

I still need to update the README.

### DIFF
--- a/src/arm-ubuntu-spec-2006/README.md
+++ b/src/arm-ubuntu-spec-2006/README.md
@@ -1,0 +1,201 @@
+---
+title: Linux arm-ubuntu image
+tags:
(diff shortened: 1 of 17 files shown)


erin-le:
I am not currently working on this, so it is being closed. I will reopen it if it becomes relevant again.

PR #54, opened 2024-09-16
resources: Update ROI

Harshil2107:
- workbegin now is called after timer_start
- workend is called before timer_end

### DIFF
--- a/src/npb-24.04-imgs/npb-with-roi/NPB/NPB3.4-OMP/LU/ssor_doac.f90
+++ b/src/npb-24.04-imgs/npb-with-roi/NPB/NPB3.4-OMP/LU/ssor_doac.f90
@@ -56,9 +56,11 @@
 
 
       call timer_start(1)
+if (niter > 1) then
 #ifdef M5_ANNOTATION
       call m5_work_begin_interface
 #endif
+endif
 !---------------------------------------------------------------------
 !   the timestep loop
 !---------------------------------------------------------------------
@@ -231,9 +233,11 @@
       end do
   900 continue
 
+if (niter > 1) then
 #ifdef M5_ANNOTATION
       call m5_work_end_interface
 #endif
+endif
 
       call timer_stop(1)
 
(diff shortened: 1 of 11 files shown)


src/npb-24.04-imgs/npb-with-roi/NPB/NPB3.4-OMP/LU/ssor_hp.f90
```diff
@@ -54,12 +54,13 @@ subroutine ssor(niter)
          call timer_clear(i)
       end do
 
+
+      call timer_start(1)
+ 
 #ifdef M5_ANNOTATION
       call m5_work_begin_interface
```
studyztp, line 62: I think both the `ssor` function in  `ssor_hp.f90` and `ssor_doac.f90` are called the same way as `ssor.f90`, which means there will be one warm up call follows by the actual call. In order to avoid calling the ROIs in the warm up call, we need to check the `niter` here just like in `ssor.f90`.
Here is a link to how `ssor_hp.f90` and `ssor_doac.f90` is compiled: https://github.com/gem5/gem5-resources/blob/stable/src/npb-24.04-imgs/npb-with-roi/NPB/NPB3.4-OMP/LU/Makefile#L24

src/npb-24.04-imgs/npb-with-roi/NPB/NPB3.4-OMP/LU/ssor_doac.f90
```diff
@@ -54,12 +54,11 @@ subroutine ssor(niter)
          call timer_clear(i)
       end do
 
+
+      call timer_start(1)
 #ifdef M5_ANNOTATION
       call m5_work_begin_interface
```
studyztp, line 61: Same as the comment in `ssor_hp.f90`

studyztp (changes requested):
Thank you for making the changes! It looks good. I only have one suggestion to the LU's ROIs.

studyztp (approved):

powerjg:
@studyztp can you re-review?

powerjg (approved):

PR #56, opened 2024-09-17
resources: Update readme to point to resources

### LINKED ISSUE #1561: Update README.md of resources in gem5-resources repo to the resources on the resources website.
Currently, the READMEs in the gem5-resources repository do not link to the resource on the resources website. It would be good to have the documentation link to the actual resources.

Harshil2107:


powerjg (commented):
This looks OK to me. The main problem will be: Will you remember to update this when you rebuild the resources?

Harshil2107:
I can update the links to be not resource version specific and so it will always point to the latest version of a resource in case we update it. If we want to make it version specific, I think would it be better to have a document that outlines what to update when a resource is updated?

powerjg:
Let's remove the version numbers.

Harshil2107:
Ok, How do we want to handle the asm tests, as we have a lot (in hundreds) and it doesn't seem helpful to have over a hundred links.

PR #58, opened 2024-10-09
resourecs: Update make.def to make is.D.x for x86 image

Harshil2107:
This PR updates the compiler flags in make.def to make is.D.x for the x86 image


src/npb-24.04-imgs/x86-npb.pkr.hcl
```diff
@@ -57,6 +57,12 @@ build {
     source      = "npb-with-roi/NPB/NPB3.4-OMP"
     destination = "/home/gem5/"
   }
+
+  provisioner "file" {
+    source      = "make-files/x86/make.def"
```
powerjg: Usually, makefile is one word without a hyphen. Unless there is a good reason to differ from the common use of "makefile", let's remove the hyphen in the directory name.

powerjg (dismissed):

erin-le (dismissed):
I don't have a good understanding of the make.def files, but I ran is.a, is.c, and is.d using the built disk image. is.a and is.c run while is.d seg faults, but this is probably expected behavior since it uses more memory than the X86Board can accommodate.

mbabaie:
I created the X86 disk image from this repo from scratch. Tested it on my own infrastructure that actually has a memory size > 3 GiB allowing to test `is.D` and it worked!

erin-le:
LGTM

erin-le (approved):
LGTM

PR #63, opened 2024-12-19
resources: Disable systemd services for x86 image

Harshil2107:


erin-le (approved):
Still have to test it, but approving so it doesn't become a blocker

erin-le:
Tested it by building and running the 22.04 and 24.04 x86 images, looks good

PR #68, opened 2025-01-22
resources: Scripts to build disk with gem5 driver

mahyarsamani:
This change adds scripts to build disk-images that include an installation of gem5-bridge kernel object so that m5 memory can be mapped without the need for sudo access. This is necessary when benchmarks should not be run with sudo permission, e.g. when using MPI runtimes. Additionally, it further breaks down the installation steps to more granular steps that are easier to understand and potentially debug.


Harshil2107:
I am closing this PR as most of the changes of this PR are already in current stable via a different PR.

PR #70, opened 2025-03-22
hacc: fix bug where numThreads were different in places

mattsinc:
nthreads and numThreads are used in different places in HACC.  However, it appears that numThreads > nthreads causes segfaults because the iq vector is accessed out of bounds.  To resolve this, create a new global variable with the number of threads and use this for both numThreads and nthreads.  I tested this worked with 4 threads.

Sidenote: although HACC refers to these as threads, they appear to actually be GPU streams.

### DIFF
--- a/src/gpu/halo-finder/src/RCBForceTree.cxx
+++ b/src/gpu/halo-finder/src/RCBForceTree.cxx
@@ -91,6 +91,8 @@ Hal Finkel (hfinkel@anl.gov)
 #include <assert.h>
 using namespace std;
 
+// use a single global variable to avoid bugs with nthreads and numThreads below
+int overallThreads = 1;
 #ifdef __HIPCC__
 #include <cudaUtil.h>
 
@@ -401,7 +403,7 @@ RCBForceTree<TDPTS>::RCBForceTree(
   vz = zVel;
   mass = ms;
 
-  numThreads=1;
+  numThreads=overallThreads;
 
   // static size for the interaction list
   #define VMAX ALIGNY(16384)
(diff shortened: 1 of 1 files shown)


src/gpu/halo-finder/src/RCBForceTree.cxx
```diff
@@ -91,6 +91,8 @@ Hal Finkel (hfinkel@anl.gov)
 #include <assert.h>
 using namespace std;
 
+// use a single global variable to avoid bugs with nthreads and numThreads below
+int overallThreads = 4;
```
powerjg: Why 4 and not 1 like in the previous code?

mattsinc: Sorry, that was a copy-paste error from applying my internal patch where I was testing with more streams, now fixed.

powerjg (approved):
@Harshil2107 can you update this binary too?

Harshil2107:
I will update the binary and upload it to the cloud bucket

mattsinc:
> I will update the binary and upload it to the cloud bucket

@Harshil2107 just checking if binary is updated and I can merge this?

Harshil2107:
I have updated the binary. The updated binarys is at: https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/halo-finder/ForceTreeTest20250424

mattsinc:
> I have updated the binary. The updated binarys is at: https://storage.googleapis.com/dist.gem5.org/dist/v24-0/test-progs/halo-finder/ForceTreeTest20250424

Thanks, I'll merge this then!

PR #71, opened 2025-05-18
resources: ROCm 6.4, Ubuntu 24.04 disk, MI300 roms

abmerop:
Bump to ROCm 6.4 and Ubuntu 24.04. These newer versions are using ACPI functionality that gem5 does not yet support. Therefore, this update also includes a Linux kernel module which implements those functions as stubs which does nothing except log an error if somehow called.

Add files to model real MI300X rather than MI210 that can run gfx942 codes. Update READMEs with new version numbers and prefer MI300X over MI210 now.

Remove TensorFlow entirely and keep same version of PyTorch. The latest versions of PyTorch for some reason run a massive amount of BLIT kernels which are slow to simulate. The ISCA'24 tutorial example goes from 3-5 minutes to 6-7 *hours* in the newest version.


abmerop:
This new disk image is required for https://github.com/gem5/gem5/pull/2284 and provides forward compatibility, but the old configs will break.

mattsinc (approved):
In general this is fine, although the increase in runtime is worrying and it would be good to understand why TensorFlow is being removed?

abmerop:
> In general this is fine, although the increase in runtime is worrying and it would be good to understand why TensorFlow is being removed?

Yea I am still trying to understand the runtime issue.  I believe the problem is the newer versions are trying to preload the entire hipBLASLt library and friends into device memory.  This of course is completely unnecessary for a short and simple simulation.  This preloading happens using BLIT kernels though, so I do have a "fix" which emulates BLIT kernels but I need to work on it more.

Tensorflow was removed partly for space reasons and because according to most of the ML community PyTorch has "won" the battle for now.

powerjg:
@Harshil2107 Please put this in a place that you won't forget... this PR got merged, but we don't have the disk created. We need to create the disk, possibly create workloads, and upload to gem5 resources. 

@mattsinc I would rather wait for merging these things until we get the binaries created. If you have a better suggestions to make sure we don't forget, I'm all ears.

mattsinc:
> @Harshil2107 Please put this in a place that you won't forget... this PR got merged, but we don't have the disk created. We need to create the disk, possibly create workloads, and upload to gem5 resources.
> 
> @mattsinc I would rather wait for merging these things until we get the binaries created. If you have a better suggestions to make sure we don't forget, I'm all ears.

Sorry ... I will just stop merging things in without explicit permission then.

PR #73, opened 2025-06-12
[riscv-fs] fix stucking at bootup

studyztp:
I found that with the ubuntu-24.04 preinstalled image, it requires one more <enter> before the login prompt.
Also, we should not redownloaded the downloaded sources. The document shows that it is a x86 image while the document is for riscv.


powerjg (commented):
lgtm. I'll let Harshil sign off.

Harshil2107 (approved):

PR #74, opened 2025-07-23
gpu: Use known-working Linux version instead of latest

abmerop:
An issue arose right after stable release where the new kernel does not load the gpu driver. Rather than using the latest kernel, fix the kernel version for release to help keep the disk build stable.


powerjg (approved):
Do we have a disk on resources.gem5.org? Does that disk work?

abmerop:
> Do we have a disk on resources.gem5.org? Does that disk work?

No, as far as I can tell there is no resource for this.

mattsinc:
sorry I didn't review this fast enough

PR #82, opened 2026-06-17
resources: Bump GPU disk image kernel

abmerop:
The GPU disk fixes a kernel version to ensure the same kernel is used when gem5 is released vs. in the future when a user might rebuild.

The kernel used in the stable GPU disk appears to have been obsoleted by Ubuntu and is no longer available. Therefore the current stable disk fails to build. Bump version to 6.8.0-124-generic from 6.8.0-79-generic. Re-ran local CI tests to confirm the new disk builds and can run applications.

### DIFF
--- a/src/x86-ubuntu-gpu-ml/scripts/rocm-install.sh
+++ b/src/x86-ubuntu-gpu-ml/scripts/rocm-install.sh
@@ -87,7 +87,7 @@ sudo chmod 777 /usr/lib/firmware/amdgpu/ip_discovery.bin
 
 # Install a known-working version of Linux as this might change after stable
 # release. Install this after DKMS so they are rebuilt.
-KERNEL=6.8.0-79-generic
+KERNEL=6.8.0-124-generic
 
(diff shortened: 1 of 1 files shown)


abmerop:
Yes, the stable branch is the intended target. I don't know if Ubuntu announces somewhere that is it obsoleting packages. I noticed this because our nightly CI started failing two days ago and just looked into why.

mattsinc:
> Yes, the stable branch is the intended target. I don't know if Ubuntu announces somewhere that is it obsoleting packages. I noticed this because our nightly CI started failing two days ago and just looked into why.

I think this will require a hotfix then?

mattsinc (approved):

abmerop:
> > Yes, the stable branch is the intended target. I don't know if Ubuntu announces somewhere that is it obsoleting packages. I noticed this because our nightly CI started failing two days ago and just looked into why.
> 
> I think this will require a hotfix then?

Not sure there are really 'hoxfixes' for resources. I don't think I have a way to change the artifacts either.

PR #83, opened 2026-08-20
RISC-V KVM resource

selimsandal:
@erin-le as requested. I don't know how these resources are hosted on the website (https://resources.gem5.org/), this resource should be uploaded there for the next release. I removed mentions of my own repo but since there is not a release that contains this resource right now it'll 404


erin-le:
Thanks @selimsandal! @BobbyRBruce will take care of uploading the resource.

powerjg (commented):
Thanks for posting this! It's great to have shared resources for everyone to be able to take advantage of these features and to be able to reproduce work that others have done.

There are some issues with this PR that need to be addressed before it's merged as an official resource.

1. Is it possible to make the init script unified between addr mode and magic instructions like the arm and x86 versions? https://github.com/gem5/gem5-resources/blob/stable/src/ubuntu-generic-diskimages/files/x86/gem5_init.sh
2. I don't want to have a patch on the m5 utility in this PR. That change should be merged into the gem5 repo.
3. We need to have the kernel module compiled for riscv. Like x86 and Arm.

PR #84, opened 2026-08-21
resources: Add MI355X GPU smoke checkpoint sources

BobbyRBruce:
## Overview

- add the HIP source and `--genco` build recipe for the
  `x86-mi355x-gpu-fs-smoke` code object
- add an initialized HIP loader and two-stage Atomic CPU configuration for
  reproducing `x86-mi355x-gpu-fs-smoke-checkpoint`
- document creation of the intermediate loader checkpoint and the final
  completed-warm-up checkpoint
- document portable checkpoint archives without macOS AppleDouble files

This PR contains only reproducible source and configuration. Generated
artifacts remain external gem5 resources. The configuration depends on the
MI355X checkpoint support in
[gem5 #3401](https://github.com/gem5/gem5/pull/3401).

## Published resources

`x86-mi355x-gpu-fs-smoke` version `1.0.0` is a 4,832-byte `gfx950` code object
with MD5 `007bb0aa417997df90195e8f2cdb00e0`.

`x86-mi355x-gpu-fs-smoke-checkpoint` version `1.0.0` is a 120,005,120-byte
portable tar archive with extracted-directory MD5
`fb11305e3e952b8764bf76f190c1360c`. Its name and metadata make clear that the
saved state contains the loaded smoke kernel and a completed warm-up dispatch;
it cannot inject a different kernel after restoration.

Both resources are published in Azure Blob Storage and the production
resource database.

## Validation

- Python source compiles and `git diff --check` passes.
- `make -n -C src/x86-mi355x-gpu-fs-smoke` selects
  `hipcc --genco --offload-arch=gfx950`.
- Both resources passed public URL/API verification and a fresh download and
  checksum validation through a built gem5 25.1 `obtain_resource`.
- The final checkpoint was created at tick `15666080342919` after a completed
  warm-up dispatch.
- A local restore launched and verified a second GPU kernel; serial output
  contained `GPU checkpoint restore test passed` and statistics recorded one
  completed workgroup.
- The archive contains no macOS AppleDouble files.


powerjg (dismissed):
CC: @abmerop and @mattsinc

powerjg (approved):

mattsinc:
@abmerop since you know this code better than I, can you please take a look first?

src/x86-mi355x-gpu-fs-smoke-checkpoint/create-checkpoint.py
```diff
@@ -0,0 +1,113 @@
...
+parser.add_argument(
+    "--checkpoint-output",
+    type=Path,
+    required=True,
+    help="Directory in which to save the generated checkpoint",
+)
+args = parser.parse_args()
+
+gem5_root = args.gem5_root.resolve()
+config = gem5_root / "tests" / "gem5" / "gpu" / "configs" / "mi355x_gpu.py"
```
abmerop, line 70: Is this file part of another PR? I wasn't able to find it

abmerop (commented):
Interesting, I don't think I have seen a python-based dlopen-like wrapper for GPU kernel libraries, but seems like this makes it easier to load the script into gem5.

High-level comment: Does it make sense for these folders to be in the gem5-resources/src/ directory? It seems like that base directory could get cluttered fast this way. Would it make more sense to put it in gem5-resources/src/gpu (or gpu-fs or some new directory)?

